PR #5, opened 2017-01-26
removed node_modules from webpack scss loader

iwwwi:
## What does this PR do:
* updates packages versions (greenkeeper updates)
* removes node_modules from scss loader in webpack because of build problems when loading the package as a npm module.

* differences

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -31,7 +31,6 @@ module.exports = {
       },
       {
         test: /\.s?css$/i,
-        exclude: /(node_modules)/,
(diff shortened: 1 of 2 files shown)


webpack.config.js
```diff
@@ -31,7 +31,6 @@ module.exports = {
       },
       {
         test: /\.s?css$/i,
-        exclude: /(node_modules)/,
```
AlexDudar: correct me if i'm wrong, if you remove `exclude`, webpack will come to node_modules and parse everything inside? is it something we would like to do to all modules? maybe we should make ui-kit as exception and but keep node_modules excluded? 

cc. @fahad19

Issue #6, opened 2017-01-28
Convert the project from a 'webpack' execution to a CLI tool


mAiNiNfEcTiOn:
I am creating this to start a discussion about it.

## Current status

The way `travix-ui-kit` is used is by triggering `webpack` from the `node_modules` (in order to use its `webpack.config.js`) to generate the base files.

Also it is receiving an env variable called `THEME_PATH` in order to specify where a theming file is located.

## Proposed change

I would actually change the `travix-ui-kit` implementation to be a CLI tool where the projects would use it on their build processes as a system to produce outputs:

```bash
npm install 'travix-ui-kit';

node_modules/.bin/travix-ui-kit --theme-file "path/to/my/theme/file.yaml" --output "generic/output/directory/for/both/jsAndCssFiles"

# or
node_modules/.bin/travix-ui-kit --theme-file "path/to/my/theme/file.yaml" --js "specific/output/directory/for/js/file/" --css "specific/output/directory/for/css/file/"

# or without theme

node_modules/.bin/travix-ui-kit --output "generic/output/directory/for/both/jsAndCssFiles"
node_modules/.bin/travix-ui-kit --js "specific/output/directory/for/js/file/" --css "specific/output/directory/for/css/file/"
```

(On `npm` _scripts_ the `node_modules/.bin/` is not needed, as you know).

This will generate the files at the given directory/directories.

Also, would mean that the usage process does not need to know where the files are or do any `copy` of a file in the `node_modules` into somewhere else.

mAiNiNfEcTiOn:
Also, continuing the proposed change above, the CLI tool could also be a module, where I would do:

```js
const uiKit = require('travix-ui-kit');

gulp.task('build:ui-kit', (done) => {
  const themeFile = 'file/path/to/theme.yaml';
  uiKit({
    output: {
      css: 'specific/output/directory/for/css/file/'
      js: 'specific/output/directory/for/js/file/',
    },
    themeFile: 'path/to/my/theme/file.yaml'
  });

  // Or if the output path is generic:
  uiKit({
    output: 'generic/output/directory/for/both/jsAndCssFiles',
    themeFile: 'path/to/my/theme/file.yaml'
  });

  // Or even without theme file:
  uiKit({
    output: 'generic/output/directory/for/both/jsAndCssFiles',
  });
});
```

asci:
I completely agree with this proposal. The way how we did it is not really perfect, it was PoC

fahad19:
[...]

PR #7, opened 2017-02-02
[WIP] Implementation of issue #6's proposal

mAiNiNfEcTiOn:
# What does this PR do:

Changes the way `travix-ui-kit` works. Currently in order to build the UI bundles we need to access the `node_modules/travix-ui-kit/` and run the `webpack` command in it in order to generate the bundles.

Also the _theming_ capability was dependent on [if an ENV var was set (`THEME_PATH`)](https://github.com/Travix-International/travix-ui-kit/blob/master/bin/build-theme.js#L5)

So, according to #6 's proposal, I created this PR to implement to concept of a builder which can also be imported via `require('travix-ui-kit/builder');`

So, usage of the CLI:

```bash
# Without watcher
node_modules/.bin/travix-ui-kit -c <pathToStoreCSS> -j <pathToStoreJS> -t <pathToThemeYAML>

# With watcher
node_modules/.bin/travix-ui-kit -c <pathToStoreCSS> -j <pathToStoreJS> -t <pathToThemeYAML> -w
```
When not provided, the theme file falls back to the `_default.yaml`.

Honestly I think the watcher is a goodie for this project's development only. Don't see a case where you want a watcher to the UI Kit in an external project.

  - Diffs
  - Clone the repo, do `npm link`, then try to use it on an existing or new code (by doing `npm link travix-ui-kit` on the folder you want to use this project on).

Still working on them (hence the WIP on the title)


bin/build-theme.js
```diff
@@ -1,22 +0,0 @@
-const path = require('path');
```
iwwwi: pls keep in mind that this file is used in the `build-theme` and `build-theme:watch` scripts. These scripts should be updated accordingly.

mAiNiNfEcTiOn: Indeed ma'am, will do!

mAiNiNfEcTiOn: @iwwwi I've refactored the code a little bit so that if we do:

```bash
node_modules/.bin/travix-ui-kit -c ./myDestForCss -j ./myDestForJs -t ./myThemeYamlFile.yml -w
```

it will detect changes on the yaml file and re-bundles the `themes.scss` which then re-triggers the bundling of the `ui-bundle.*` files.

mAiNiNfEcTiOn: Now I am changing the `package.json` to basically use the builder :)

mAiNiNfEcTiOn:
Working on the tests, bear with me on this one :D

mAiNiNfEcTiOn:
Almost...there.

README.md
```diff
@@ -7,10 +7,66 @@ Travix UI Components' repository.
...
+```
+
+
+For simplicity purposes we suggest to add a task/script to your `package.json`,
+which simplifies the usage of the CLI. E.g.:
+
+```js
+{
+  "scripts": {
+    "build:ui": "travix-ui-kit -j ./js/ -c ./css/ -t ./myDefaultStyle.yml",
```
iwwwi: Opppsss... I am blind 🙈

README.md
```diff
@@ -7,10 +7,66 @@ Travix UI Components' repository.
...
+
+```bash
+$ node_modules/.bin/travix-ui-kit -h
+
+  Usage: travix-ui-kit [options]
+
+  Options:
+
+    -h, --help                 output usage information
+    -V, --version              output the version number
```
asci: Why is `V` capital?

mAiNiNfEcTiOn: It comes from `commander`: https://github.com/tj/commander.js/issues/560

Nevertheless it is also more common to do `-v` to add verbosity on node CLI tools. Still, it is possible to provide an override of the help I think.

README.md
```diff
@@ -22,14 +78,16 @@ Travix UI Components' repository.
...
 - or inject it in your page current styles bundle
 - or copy to public folder and add as separate style file with `<link>`
 
 **Warning**: Directly using file `components/index.scss` not recommended. We're not promising that we will use SCSS in future or will keep file's structure
 
+---
+
 ## Living style guide
 
 ### Before installation
```
iwwwi: I can not add comment on the exact line (96), since that part of the doc is not changed, but here we describe how developers of the travix-ui-kit can set a path to a different than the default theme.  This should be updated and also to be explained how to use the builder itself in this case. What i used during the testing was: `$ node ./builder -t ./themes/newTheme.yaml`.

fahad19:
💖

PR #13, opened 2017-02-13
Updated mods helper and class names according to the new BEM standards

AlexDudar:
Updated according to the new BEM standards (see [here](https://en.bem.info/methodology/quick-start/#modifier))

Modified:
- mods helper.js 
- modifiers class names
- unit tests

Modifiers were changed from `block-name--modifier` to `block-name_modifier`

### DIFF
--- a/components/_helpers.js
+++ b/components/_helpers.js
@@ -1,4 +1,4 @@
-const modSep = '--';
+const modSep = '_';
 
 function getClassNamesWithMods(baseClass, mods = []) {
   baseClass = [].concat(baseClass);
--- a/components/button/button.scss
+++ b/components/button/button.scss
@@ -23,7 +23,7 @@ $text-shadow: null !default;
   transition: all .1s ease-in-out;
 }
 
-.ui-button--variation_default {
+.ui-button_variation_default {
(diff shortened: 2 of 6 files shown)


components/button/button.js
```diff
@@ -16,15 +16,15 @@ function Button({ children, mods = [], size, href, onClick, type, variation, dis
   }
 
   if (size) {
-    mods.push(`size_${size}`);
+    mods.push(`size-${size}`);
```
asci: According to BEM naming convention the dash `-` could be used only for breaking words in long names, but not for separation values

AlexDudar: yes, as we discussed yesterday, there is this possibility. I think we can do this, but let's all align, @mAiNiNfEcTiOn @iwwwi what do you think? Actually which one do you prefer `ui-button_size-s` or `ui-button_size_s` ?

asci: just as an additional example (not in button, but for some future components):
`app-bar_show-icon-left` or `app-bar_show-icon_left`?

AlexDudar:
@asci I made adjustments according to your comment, please check

PR #16, opened 2017-02-15
Added List component

AlexDudar:
Added List component (later on will be extended with possibility to have icon)

### DIFF
--- a/components/list/list.js
+++ b/components/list/list.js
@@ -1,0 +1,60 @@
+// Imports
+import React, { PropTypes } from 'react';
+import { getClassNamesWithMods } from '../_helpers';
+
+/**
+ * General List component. Use when you need to display array of elements
+ */
+function List({ items, hideBullets, mods = [], orientation }) {
+  mods.push(`orientation_${orientation}`);
+
+  if (hideBullets) {
+    mods.push("no-bullets");
+  }
+
+  const className = getClassNamesWithMods('ui-list', mods);
+
+  const itemsBlock = items.map((item, index) => (
+    <li key={index}>
+      <span className="ui-list__content">{item}</span>
+    </li>
+  ));
+
+  return (
+    <ul className={className}>
+      {itemsBlock}
+    </ul>
+  );
+}
+
+List.defaultProps = {
+  hideBullets: false,
+  orientation: 'vertical',
+};
+
+List.propTypes = {
+  /**
+   * List's elements.
+   */
+  items: PropTypes.arrayOf(PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.element,
+  ])).isRequired,
+
+  /**
+   * Hide list's bullets
+   */
+  hideBullets: PropTypes.bool,
+
+  /**
+   * You can provide set of custom modifications.
+   */
+  mods: PropTypes.arrayOf(PropTypes.string),
+
+  /**
+   * List's apperance.
+   */
+  orientation: PropTypes.oneOf(['vertical', 'horizontal']),
+};
+
+export default List;
--- a/components/list/list.js
+++ b/components/list/list.js
@@ -5,8 +5,8 @@ import { getClassNamesWithMods } from '../_helpers';
 /**
  * General List component. Use when you need to display array of elements
  */
-function List({ items, hideBullets, mods = [], orientation }) {
-  mods.push(`orientation_${orientation}`);
+function List({ items, hideBullets, mods = [], align }) {
+  mods.push(`align_${align}`);
 
   if (hideBullets) {
     mods.push("no-bullets");
@@ -15,8 +15,8 @@ function List({ items, hideBullets, mods = [], orientation }) {
   const className = getClassNamesWithMods('ui-list', mods);
 
   const itemsBlock = items.map((item, index) => (
-    <li key={index}>
-      <span className="ui-list__content">{item}</span>
+    <li className="ui-list__item" key={index}>
+      {item}
     </li>
   ));
 
@@ -29,7 +29,7 @@ function List({ items, hideBullets, mods = [], orientation }) {
 
 List.defaultProps = {
   hideBullets: false,
-  orientation: 'vertical',
+  align: 'vertical',
 };
 
 List.propTypes = {
@@ -54,7 +54,7 @@ List.propTypes = {
   /**
    * List's apperance.
    */
-  orientation: PropTypes.oneOf(['vertical', 'horizontal']),
+  align: PropTypes.oneOf(['vertical', 'horizontal']),
 };
 
 export default List;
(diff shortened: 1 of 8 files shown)


components/list/list.js
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * You can provide set of custom modifications.
+   */
+  mods: PropTypes.arrayOf(PropTypes.string),
+
+  /**
+   * List's apperance.
+   */
+  position: PropTypes.oneOf(['vertical', 'horizontal']),
```
asci: What do you think about renaming this to 'direction'?

components/list/list.js
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * General List component. Use when you need to display array of elements
+ */
+function List({ items, hideBullets, mods = [], position }) {
+  if (position) {
+    mods.push(`position_${position}`);
+  }
+
+  if (hideBullets) {
+    mods.push("no_bullets");
```
asci: why `_` and don't `-`?

components/list/list.js
```diff
@@ -0,0 +1,59 @@
+// Imports
+import React, { PropTypes } from 'react';
+import { getClassNamesWithMods } from '../_helpers';
+
+/**
+ * General List component. Use when you need to display array of elements
+ */
+function List({ items, hideBullets, mods = [], position }) {
+  if (position) {
```
asci: I guess `if` is redundant

components/list/list.js
```diff
@@ -0,0 +1,59 @@
...
+List.defaultProps = {
+  hideBullets: false,
+  position: 'vertical',
+};
+
+List.propTypes = {
+  /**
+   * List's elements.
+   */
+  items: PropTypes.arrayOf(PropTypes.string).isRequired,
```
asci: would be also nice to add other components as items, for example `<Text html>` would be useful if you need to have links in list items

components/list/list.js
mAiNiNfEcTiOn: or maybe orientation 🙄 ...

components/list/list.js
AlexDudar: couldn't decide so just sticked to _ ..can be `no-bullets` since `bullets` is not a value, but part of modifier name

components/list/list.js
AlexDudar: i would go with `orientation`

components/list/list.js
AlexDudar: well, I can make it like this:
`
items: PropTypes.oneOfType([
    PropTypes.arrayOf(PropTypes.string),
    PropTypes.element,
  ]).isRequired,
`

so we can pass an element, but why do you need to use List component to pass one Text or Button there? or you are going to make an array of components? sorry, but maybe i'm not getting your idea. 

in case if we need to have list's elements wrapped with `a` or `span` we can make a prop `type` where we can specify if it should be a link or something else

AlexDudar: I had a chat with Ivana so now it's clear what do you want to do. I will update my PR soon

components/list/list.scss
```diff
@@ -0,0 +1,31 @@
+@import "../../themes/theme";
+
+.ui-list {
+  font-family: $tx-generic-font-secondary-font-family, $tx-generic-font-secondary-generic-family;
+
+  &.ui-list_orientation_vertical li {
```
ibsukru: http://thesassway.com/intermediate/avoid-nested-selectors-for-more-modular-css

components/list/list.js
```diff
@@ -0,0 +1,60 @@
...
+  return (
+    <ul className={className}>
+      {itemsBlock}
+    </ul>
+  );
+}
+
+List.defaultProps = {
+  hideBullets: false,
+  orientation: 'vertical',
```
ibsukru: **align** is better word maybe, orientation is confusing with orientation:portrait, orientation:landscape

components/list/list.scss
```diff
@@ -0,0 +1,31 @@
...
+    color: $tx-list-bullets-color;
+    line-height: 38px;
+  }
+
+  &.ui-list_orientation_horizontal {
+    padding: 0;
+
+    li {
```
ibsukru: http://thesassway.com/intermediate/avoid-nested-selectors-for-more-modular-css

components/list/list.js
```diff
@@ -0,0 +1,60 @@
...
+
+  if (hideBullets) {
+    mods.push("no-bullets");
+  }
+
+  const className = getClassNamesWithMods('ui-list', mods);
+
+  const itemsBlock = items.map((item, index) => (
+    <li key={index}>
+      <span className="ui-list__content">{item}</span>
```
ibsukru: why we wrapped with span here ?

AlexDudar: to make list bullets and text in different colors

ibsukru: li:before {content: "•"; color: red;
  display: inline-block; width: 1em;
  margin-left: -1em}

components/list/list.scss
```diff
@@ -3,15 +3,27 @@
...
 
-  &.ui-list_orientation_vertical li {
-    color: $tx-list-bullets-color;
-    line-height: 38px;
+  &.ui-list_align_vertical {
+    .ui-list__item {
+      display: block;
+      line-height: 38px;
+
+      &:before {
```
ibsukru: looks good 👍

AlexDudar:
@iwwwi please merge my PR, already rebased, and approved by people :)

components/list/list.js
```diff
@@ -0,0 +1,60 @@
+// Imports
+import React, { PropTypes } from 'react';
+import { getClassNamesWithMods } from '../_helpers';
+
+/**
+ * General List component. Use when you need to display array of elements
+ */
+function List({ items, hideBullets, mods = [], align }) {
+  mods.push(`align_${align}`);
```
fahad19: this would have a side-effect. especially if the `mods` array is passed as a reference from the top-level.

just saying 😬

fahad19: just looked at the `master` branch. I see other components doing the same too. At some point they all need to be updated.

I remember dealing with a similar reference issue (by passing props in a loop) with @iwwwi last year.

asci: good point, we need to fix it I guess

iwwwi: @fahad19 (_correct me if I am wrong_ ) As I remember the problem was when having sibling components and passing object/array their rendering was affected from their sister components. Also what i remember at that time the problem was solved by defining default empty array/object for each component and then populating it with elements/properties.

AlexDudar: @asci @iwwwi @fahad19 guys, I don't really know what you were discussing before, but could you please catch up and decide something on actions, so we can move on with this PR

fahad19: I am happy to quickly discuss it tomorrow.

@AlexDudar: I think you can merge the PR as is. the reference issue can be taken care of later along with all the other components.

PR #18, opened 2017-02-17
Price component

### LINKED ISSUE #11: Price component

mAiNiNfEcTiOn:
# What does this PR do:
Adds the Price component to Travix UI Kit.

Fixes #11.

screen shot 2017-02-22 at 17 11 35

  - Diffs
  - Look at the styleguide

Adds Unit tests for Price component, reaching 100% coverage

### DIFF
--- a/components/price/price.js
+++ b/components/price/price.js
@@ -10,7 +10,7 @@ const { PropTypes } = React;
  * it will return the original value.
  *
  * @function addThousandsSeparator
- * @param {String} value              Integer value to be formatted
+ * @param {String} value              String to be formatted
  * @param {String} thousandsSeparator Character to be used to split the thousands unit.
  * @return {String}                   Either the original value or the formatted value.
  */
@@ -31,7 +31,7 @@ export function addThousandsSeparator(value, thousandsSeparator) {
  * Ensures that a decimal value has the right decimal precision (no rounding applied).
  *
  * @function ensureDecimalPrecision
- * @param {String} value              Integer value, representing the decimal part of a number.
+ * @param {String} value              String representing the decimal part of a number.
  * @param {Number} decimalsPrecision  Number of decimals to be allowed.
  * @return {String}                   The value with a proper precision.
  */
--- a/components/price/price.js
+++ b/components/price/price.js
@@ -1,8 +1,47 @@
 // Imports
 import React from 'react';
-import { getClassNamesWithMods } from '../_helpers';
+import { getClassNamesWithMods, getDataAttributes } from '../_helpers';
 
 const { PropTypes } = React;
+
+/**
+ * Adds the thousands separator to a given value.
+ * If either the thousands separator is not set or the value doesn't have thousands,
+ * it will return the original value.
+ *
+ * @function addThousandsSeparator
+ * @param {String} value              Integer value to be formatted
+ * @param {String} thousandsSeparator Character to be used to split the thousands unit.
+ * @return {String}                   Either the original value or the formatted value.
+ */
+export function addThousandsSeparator(value, thousandsSeparator) {
+  if (!thousandsSeparator || (value.length <= 3)) {
+    return value;
+  }
+
+  return value
+    .split('')
+    .reverse()
+    .reduce((ret, digit, idx) => ret.concat([((idx > 0) && ((idx % 3) === 0) ? thousandsSeparator : ''), digit]), [])
+    .reverse()
+    .join('');
+}
+
+/**
+ * Ensures that a decimal value has the right decimal precision (no rounding applied).
+ *
+ * @function ensureDecimalPrecision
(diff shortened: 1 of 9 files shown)


asci:
can you please provide a picture how it is looks like in LSG?

components/price/price.js
```diff
@@ -0,0 +1,114 @@
...
+  const rootClass = 'ui-price';
+  const className = getClassNamesWithMods(rootClass, mods);
+  const [intValue, decValue] = value.toString().split('.');
+
+  let formattedIntValue = intValue;
+  if (thousandsSeparator && (formattedIntValue.length > 3)) {
+    formattedIntValue = formattedIntValue
+      .split('')
+      .reverse()
+      .reduce((ret, digit, idx) => ret.concat([((idx > 0) && ((idx % 3) === 0) ? thousandsSeparator : ''), digit]), [])
```
asci: what do you think about extract this function? for example, name it 'removeThousandsSeparator'. IMHO would be better for readability.

components/price/price.js
```diff
@@ -0,0 +1,114 @@
...
+      .reverse()
+      .join('');
+  }
+
+  const formattedDecValue = decValue.length > decimalsPrecision
+    ? decValue.substr(0, decimalsPrecision)
+    : `${decValue}${'0'.repeat(decimalsPrecision - decValue.length)}`;
+
+  const dataAttributes = Object.keys(dataAttrs).reduce((ret, key) => {
```
asci: what about moving it to the `_helpers.js` file?

mAiNiNfEcTiOn: Will do, please stand-by :)

components/price/price.js
mAiNiNfEcTiOn: 👍  Will do so...

components/price/price.js
```diff
@@ -0,0 +1,145 @@
...
+
+const { PropTypes } = React;
+
+/**
+ * Adds the thousands separator to a given value.
+ * If either the thousands separator is not set or the value doesn't have thousands,
+ * it will return the original value.
+ *
+ * @function addThousandsSeparator
+ * @param {String} value              Integer value to be formatted
```
asci: it is not Integer

asci:
Do we have only one size for Price component?

mAiNiNfEcTiOn:
@asci again, the breakpoints are not supposed to be in the `travix-ui-kit`.

1. Not everyone has the same breakpoints
1. Not everyone wants to do responsive web design
1. Would not be flexible enough....

I would rather leave that for the a specific theme where you'd do something like:

```scss
.ui-price {
  .ui-price__currency {
    @include breakpoint('lt-medium') {
      font-size: 15px;
    }
  }
}
```

Or, if possible, override the variables would be even better.

components/price/price.js
```diff
@@ -0,0 +1,145 @@
...
+  const rootClass = 'ui-price';
+  const className = getClassNamesWithMods(rootClass, mods);
+  const [intValue, decValue] = value.toString().split('.');
+
+  const decimalsMarkup = showDecimals ? (<div className={`${rootClass}__decimals`}>{
+    decimalsSeparator + ensureDecimalPrecision(decValue, decimalsPrecision)
+  }</div>) : null;
+  const underlineMarkup = underlined ? <div className={`${rootClass}__underline`}/> : null;
+
+  return (<div className={className} {...getDataAttributes(dataAttrs)}>
```
AlexDudar: I would put html to the next line (outside of parentheses), to increase code readability

components/price/price.scss
```diff
@@ -0,0 +1,55 @@
...
+  flex-flow: row nowrap;
+  flex: 0 0 auto;
+}
+
+.ui-price__currency {
+  display: flex;
+  font-size: $tx-price-font-size-currency;
+  flex: 0 0 auto;
+
+  &.ui-price__currency-left {
```
AlexDudar: what is `left` in context of this class? is it modifier or element's value?

asci:
@mAiNiNfEcTiOn I'm not talking about responsiveness, I'm talking about semantics

components/price/price.scss
mAiNiNfEcTiOn: Well it's both a modifier and an element's value... But I'd go for modifier...

Should be separated by `_`, is that?

mAiNiNfEcTiOn:
@asci define _semantics_ and what you're expecting and/or proposing please.

components/price/price.js
mAiNiNfEcTiOn: I'm going to move it to the next line, but inside of the parenthesis :)

asci:
@mAiNiNfEcTiOn  I mean we have different sizes for a button (small, medium, large). Why don't we have it for a price component? Maybe we have different sizes in design?

PR #25, opened 2017-02-23
Radio button

iwwwi:
## What does this PR do:

* fixes issue #17
* adds stateless radio button component.
* adds custom styles to  the radio button.
* adds aria properties for the radio button since we are not using the native html tag
* adds test
* adds md
* updates snapshots

* components/radioButton/radioButton.js
* components/radioButton/radioButton.scss
* components/radioButton/radioButton.md

screen shot 2017-02-23 at 11 22 22

### DIFF
--- a/components/radioButton/radioButton.js
+++ b/components/radioButton/radioButton.js
@@ -44,6 +44,11 @@ RadioButton.defaultProps = {
 
 RadioButton.propTypes = {
   /**
+   * Represents the state of the radio button.
+   */
+  checked: PropTypes.bool,
+
+  /**
    * Radio Button value.
    */
   children: PropTypes.oneOfType([
@@ -53,12 +58,7 @@ RadioButton.propTypes = {
   ]),
 
   /**
-   * Represents the state of the radio button.
-   */
-  checked: PropTypes.bool,
-
-  /**
-   * Data attribute. You can use it to set up GTM key or any custom data-* attribute.
+   * Data attribute. You can use it to set up any custom data-* attribute.
    */
   dataAttrs: PropTypes.oneOfType([
     PropTypes.bool,
--- a/components/radioButton/radioButton.js
+++ b/components/radioButton/radioButton.js
@@ -1,0 +1,94 @@
+// Imports
+import React, { PropTypes } from 'react';
+
+import { getClassNamesWithMods, getDataAttributes } from '../_helpers';
+
+function RadioButton({
+  checked,
+  children,
+  dataAttrs = false,
+  disabled,
+  id,
+  mods = [],
+  name,
+  onChange }) {
+  const restProps = getDataAttributes(dataAttrs);
+
+  if (disabled) {
+    mods.push('disabled_true');
+  }
+
+  const className = getClassNamesWithMods('ui-radio', mods);
+
+  return (
+    <div className={className} {...restProps}>
+      <input
+        checked={checked}
+        disabled={disabled}
+        id={id}
+        name={name}
+        onChange={onChange}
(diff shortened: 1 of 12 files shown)


components/radioButton/radioButton.js
```diff
@@ -0,0 +1,94 @@
...
+    <div className={className} {...restProps}>
+      <input
+        checked={checked}
+        disabled={disabled}
+        id={id}
+        name={name}
+        onChange={onChange}
+        type="radio"
+      />
+      <label aria-checked={checked} htmlFor={id} role="radio">
```
ibsukru: You don't need aria-checked={checked} | role="radio", unless you are using for='radioId'

iwwwi: Here we are using the label to also style the input element. which means we are loosing the accessibility properties from the native input element. In this case we need to add this props so that we can navigate as we have the native element (ex.: navigating with tabs and focusing the element).

components/radioButton/radioButton.js
```diff
@@ -0,0 +1,94 @@
...
+};
+
+RadioButton.propTypes = {
+  /**
+   * Represents the state of the radio button.
+   */
+  checked: PropTypes.bool,
+
+  /**
+   * Radio Button value.
```
asci: Radio Button __label__

components/radioButton/radioButton.scss
```diff
@@ -0,0 +1,61 @@
...
+  font-family: $tx-generic-font-secondary-font-family, $tx-generic-font-secondary-generic-family;
+  font-size: 14px;
+  font-weight: $tx-generic-font-primary-weight-regular;
+  min-height: 20px;
+  padding-bottom: 2px;
+  padding-right: 10px;
+  padding-top: 2px;
+  position: relative;
+
+  & input[type="radio"] {
```
asci: is this `input[type="radio"]`required?

components/radioButton/radioButton.js
ibsukru: That is exactly what i mean :) you have label with for='radio' attribute already, aria-checked={checked} role="radio" is not required, so accessibility devices understand this is a label for radio input

mAiNiNfEcTiOn:
@iwwwi you should `git rebase master` and `npm run update-snapshots`.

components/radioButton/radioButton.scss
```diff
@@ -0,0 +1,61 @@
...
+    left: -9999px;
+    position: absolute;
+
+    & + label {
```
mAiNiNfEcTiOn: So you're saying that she should do:

```scss
.ui-price input[type="radio"] + label {
  // code here
}
```

?

ibsukru: Actually .ui-price__label { ... 
adding that class is a solution

themes/_default.yaml
```diff
@@ -40,7 +40,7 @@ generic:
     secondary-lighter: &secondary-lighter '#FFD97C'
 
     text-dark: &text-dark '#333'
-    text: &text '#666'
```
iwwwi: will undo the change :)

components/radioButton/radioButton.scss
iwwwi: @ibsukru how do you suggest to use the pseudo selectors using the  mentioned class? I am bit concerned because  i need to combine multiple elements with their pseudo selectors, for ex. `input[type="radio"]:checked + label:after`. In this case I will have to write something like `input[type="radio"]:checked + label.ui-radio__label:after`

Issue #31, opened 2017-03-09
When SASS files have errors the error is swallowed


mAiNiNfEcTiOn:
# What's the issue:

When a SASS file has an error, it's error is swallowed (and never shown), only possible to detect due to the non-existence of the `ui-bundle.css`.

<img width="778" alt="screen shot 2017-03-09 at 10 57 38" src="https://cloud.githubusercontent.com/assets/1002056/23745193/a579ed9e-04b7-11e7-8435-e508696d6397.png">
<img width="808" alt="screen shot 2017-03-09 at 10 59 14" src="https://cloud.githubusercontent.com/assets/1002056/23745192/a56be082-04b7-11e7-828e-d9a3eeeaa708.png">

PR #32, opened 2017-03-09
Fixes issue #31 - Changes the builder to properly catch rejections and to output the errors on the webpack process

mAiNiNfEcTiOn:
# What does this PR do:

* Increases the coverage by removing certain conditionals that verified if a property is set when it they have a default value.

* Changes the `builder.js` to return the `Promise` instead of executing it.

* Adds the `bail` property to the `webpack.config.js` to force it to fail immediately so that we get the exact error instead of the last one.

screen shot 2017-03-09 at 11 05 04

  - Diffs

All tests are passing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,12 +23,13 @@ $ node_modules/.bin/travix-ui-kit -h
 
   Options:
 
-    -h, --help                 output usage information
-    -V, --version              output the version number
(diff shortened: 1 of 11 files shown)


mAiNiNfEcTiOn:
Will attempt to increase coverage again :/

PR #42, opened 2017-03-16
fix for price component

maximuk:
# What does this PR do:
When an integer number is provided to the Price component, the `ensureDecimalPrecision` function does not handle the _undefined_ (attempts to convert it to string).

This PR changes the `ensureDecimalPrecision` to have a default value on the `value` attribute to be empty string, fixing the issue.

  - Diffs, specially the file `price.js`.

Tests covering the issue added.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "travix-ui-kit",
-  "version": "0.3.3",
+  "version": "0.3.2",
   "description": "Travix UI kit",
   "main": "lib/index.js",
   "scripts": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "travix-ui-kit",
-  "version": "0.3.2",
+  "version": "0.3.3",
```
iwwwi: why are you changing the version?

iwwwi:
the test coverage is not 100%. 1 test is still failing.

mAiNiNfEcTiOn:
@maximuk thanks 👍

PR #44, opened 2017-03-16
Collapse component

EduardTrutsyk:
### What does this PR do:
Adds the Collapse component to Travix UI Kit.

screenshot at 19-00-45

### Where should the reviewer start:

- Diffs
- Look at the styleguide

### Unit and/or functional tests:

Adds Unit tests for Modal component, reaching 100% coverage


components/collapse/collapse.js
```diff
@@ -0,0 +1,58 @@
...
+   */
+  children: PropTypes.node,
+  /**
+   * Specify a function that will be called when a user click on CollapseItem
+   */
+  onChange: PropTypes.func,
+};
+Collapse.defaultProps = {
+  children: null,
+  accordion: false,
```
AlexDudar: please sort

components/collapse/collapse.js
```diff
@@ -0,0 +1,58 @@
+import React, { Component, PropTypes } from 'react';
+
+import CollapseItem from './collapseItem';
```
AlexDudar: why name is Collapse not Accordion?

tests/unit/collapse/collapse.render.spec.js
```diff
@@ -0,0 +1,85 @@
+
```
AlexDudar: can be removed ;)

components/collapse/collapse.js
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * The list of CollapseItem components
+   */
+  children: PropTypes.node,
+  /**
+   * Specify a function that will be called when a user click on CollapseItem
+   */
+  onChange: PropTypes.func,
+};
+Collapse.defaultProps = {
```
iwwwi: please add empty line above

components/index.js
```diff
@@ -4,9 +4,13 @@ import Modal from './modal/modal';
 import Price from './price/price';
 import RadioButton from './radioButton/radioButton';
 import Spinner from './spinner/spinner';
+import Collapse from './collapse/collapse';
```
iwwwi: pls sort

components/collapse/collapse.js
```diff
@@ -0,0 +1,58 @@
+import React, { Component, PropTypes } from 'react';
```
asci: `, Children }`

components/collapse/collapse.js
```diff
@@ -0,0 +1,58 @@
...
+
+import CollapseItem from './collapseItem';
+
+/**
+ * Collapse component
+ */
+class Collapse extends Component {
```
asci: could be replaced by stateless component

components/collapse/collapse.js
EduardTrutsyk: accordion - it's behavior, in this case only one item can be expanded, I added props for that.

mAiNiNfEcTiOn:
@EduardTrutsyk please take a look at this article and add proper accessibility tags: http://heydonworks.com/practical_aria_examples/#progressive-collapsibles

components/collapse/collapse.js
```diff
@@ -0,0 +1,129 @@
...
+  }
+}
+
+Collapse.propTypes = {
+  /**
+   * Accordion mode, only one panel can be expanded at a time.
+   */
+  isAccordion: PropTypes.bool,
+
+  activeKey: PropTypes.oneOfType([
```
asci: missing description

components/global/global.js
```diff
@@ -1,5 +1,8 @@
 import React, { Component, PropTypes } from 'react';
-import { render, unmountComponentAtNode } from 'react-dom';
+import {
+  unstable_renderSubtreeIntoContainer, // eslint-disable-line
```
asci: why unstable? what is wrong with `render`?

tests/unit/calendar/__snapshots__/calendar.normal.spec.js.snap
```diff
@@ -591,7 +591,7 @@ exports[`Calendar (normal mode) #render() should apply the new minDate on Calend
           onNavNextMonth={[Function]}
           onNavPreviousMonth={[Function]}
           onSelectDay={[Function]}
-          renderDate={2017-03-21T00:00:00.000Z}
+          renderDate={2017-03-23T00:00:00.000Z}
```
asci: its because of timezones?

mAiNiNfEcTiOn: No... it's because of the current date... He suggested mocking the `new Date()` which IMHO leads to further complications.

I will be taking a look at it in a few minutes.

tests/unit/helpers/helpers.spec.js
```diff
@@ -12,6 +12,9 @@ describe('helpers', () => {
     it('#3 should return only an array with the classes provided in the first arg', () => {
       expect(getClassNamesWithMods(['test', 'test2'])).toBe('test test2');
     });
+    it('#4 should return empty string when baseClass was not provided', () => {
+      expect(getClassNamesWithMods()).toBe('');
+    });
```
iwwwi: What about the scenario where you pass Object instead of Array?

components/global/global.js
EduardTrutsyk: because, in case, when we will use `frint`, `redux`, or something like that, we have `Provider`, which provide `store` or another stuff, it donesn't work as expected.

because `render` will render content in to `body`, this content has no access to `Provider`, and component which connected to store.

`unstable_renderSubtreeIntoContainer` will work like `pipeline`, and we can render content into `target`. 
something like that ;)

tests/unit/calendar/__snapshots__/calendar.normal.spec.js.snap
EduardTrutsyk: no, it's because `new Date()` was not stub for Calendar's unit tests.
I created issue for that: https://github.com/Travix-International/travix-ui-kit/issues/48

components/collapse/collapse.md
```diff
@@ -0,0 +1,51 @@
...
+        <CollapseItem title="Collapse Title 2">
+          <List items={['London', 'Amsterdam', 'Madrid']} />
+        </CollapseItem>
+      </Collapse>
+    </div>
+
+Open one with expanded first item by default:
+
+    <div style={{marginTop: '10px'}}>
+      <Collapse isAccordion onChange={key => setState({ example1ActiveKey: key })} activeKey={(state.example1ActiveKey === undefined ? 'example1.1' : state.example1ActiveKey)}>
```
mAiNiNfEcTiOn: Where is this `setState()` coming from?

components/collapse/collapse.js
```diff
@@ -0,0 +1,129 @@
...
+        activeKey.splice(index, 1);
+      } else {
+        activeKey.push(key);
+      }
+    }
+    this.setActiveKey(activeKey);
+  }
+
+  setActiveKey(activeKey) {
+    if (!('activeKey' in this.props)) {
```
EduardTrutsyk: when developer will not provide it, then component will use own state, in case, when developer provide `activeKey`(from redux store for example), then collapse will update the state based on props

components/collapse/collapse.md
EduardTrutsyk: line 17: `onChange={key => setState({ example1ActiveKey: key })}`

components/collapse/collapse.js
```diff
@@ -0,0 +1,129 @@
...
+  Children,
+  cloneElement,
+  PropTypes,
+} from 'react';
+
+import { getClassNamesWithMods } from '../_helpers';
+import CollapseItem from './collapseItem';
+
+const getNormalizedActiveKey = ({ activeKey = [] }) => {
+  return activeKey instanceof Array ? activeKey : [activeKey];
```
EduardTrutsyk: I didn't change anything, why I need to create a new array? this function will return array always

components/collapse/collapse.js
```diff
@@ -0,0 +1,129 @@
...
+
+  componentWillReceiveProps(nextProps) {
+    if ('activeKey' in nextProps) {
+      this.setState({
+        activeKey: getNormalizedActiveKey(nextProps),
+      });
+    }
+  }
+
+  handleItemClick = (e, key) => {
```
EduardTrutsyk: if I will use `=>`, [babel](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-0&targets=&browsers=&builtIns=false&code=class%20C1%20%7B%0A%20%20f1()%20%7B%0A%20%20%20%20this.a%20%3D%201%3B%0A%20%20%7D%0A%20%20%0A%20%20f2%20%3D%20()%20%3D%3E%20%7B%0A%20%20%20%20this.a%20%3D%201%3B%0A%20%20%7D%0A%7D) move it to constructor, and create new function for each instance, this is the same if we do `bind` in constructor.

Issue #36, opened 2017-03-13
Calendar


mAiNiNfEcTiOn:
The calendar component must meet the following requirements (WIP, and not in the specified order):

- [x] Must be possible to define `minDate` and `maxDate` to define boundaries of selection.
- [x] Must be possible to define `onSelectDay` to pass a handler of the selected date.
- [x] Must be possible to define `isDaySelectableFn` to pass a function that should return a `Boolean` value to determine if the day is selectable.
- [x] Must be possible to define `locale`, which will receive an object with the following things:
```javascript
{
  months: [
    { name: 'January', short: 'Jan' },
    { name: 'February', short: 'Feb' },
    // ...
  ],
  startWeekDay: 1, // retrieved via new Date().getDay(), where 0 = Sunday, 6 = Saturday
  weekDays: [
    { name: 'Monday', 'short: Mon' },
  ],
}
```

PR #45, opened 2017-03-17
Calendar component - Issue #36 (WIP)

mAiNiNfEcTiOn:
# What does this PR do:

* Adds a new helper function that _normalizes_ a `Date` object's hours, minutes, seconds and milliseconds.
* Adds the Calendar component to the kit.
* Changes the `_default.yaml` to contain a basic default style of the calendar.
* Changes the builder to ouput errors via `util.inspect()` which colorizes the output as well
* Configures the rule `jsx-indent` on the `.eslintrc` to enable the `indentLogicalExpressions`, overriding its default behavior to the one of v6.10.0: https://github.com/yannickcr/eslint-plugin-react/issues/1117#issuecomment-287881398
* Locks the snapshots and tests to run on UTC (using the TZ env variable) so that it runs in the same timezone as Travis CI.

Documentation also is a WIP... Won't merge without having it properly set.

screen shot 2017-03-17 at 11 46 49
screen shot 2017-03-17 at 11 46 17
screen shot 2017-03-17 at 11 45 20
screen shot 2017-03-17 at 11 39 03

  - `components/calendar.js`

Being added (WIP). But you can start the code review without them :)


components/calendar/calendar.js
```diff
@@ -0,0 +1,268 @@
+import React from 'react';
```
asci: why not 
`import React, { PropTypes, Component } from 'react';`?

mAiNiNfEcTiOn: Indeed... would make it more readable... will do 👍

components/calendar/calendar.js
```diff
@@ -0,0 +1,268 @@
...
+const { PropTypes } = React;
+
+export default class Calendar extends React.Component {
+  constructor(props) {
+    super();
+
+    const { initialDates, maxDate, minDate } = props;
+    const maxLimit = maxDate ? new Date(maxDate) : null;
+    const renderDate = (initialDates && initialDates.length) ? new Date(initialDates[0]) : new Date();
+    const selectedDates = initialDates ? initialDates.map(item => (item ? new Date(item) : null)) : [null, null];
```
asci: two inlined ternary operators with arrow function - hard to read\understand, please consider simplifying and\or extract

components/calendar/calendar.js
```diff
@@ -0,0 +1,268 @@
...
+  CALENDAR_SELECTION_TYPE_RANGE,
+} = calendarConstants;
+
+const { PropTypes } = React;
+
+export default class Calendar extends React.Component {
+  constructor(props) {
+    super();
+
+    const { initialDates, maxDate, minDate } = props;
```
asci: is it possible by any chance that props will be changed during the lifetime of the component? if so, should you handle it somehow by `componentWillReceiveProps`?

mAiNiNfEcTiOn: @asci will add a test for that 👍

components/calendar/calendar.js
mAiNiNfEcTiOn: Sure, no problem. 👍

components/calendar/calendar.js
```diff
@@ -0,0 +1,268 @@
...
+          selectionType={selectionType}
+        />
+      </div>
+    );
+  }
+}
+
+Calendar.defaultProps = {
+  dataAttrs: {},
+  mods: [],
```
asci: potential reference issue

components/calendar/panels/days.js
```diff
@@ -0,0 +1,28 @@
+import React, { PropTypes } from 'react';
+import { getClassNamesWithMods } from '../../_helpers';
+import DaysView from '../views/days';
+
+export default class Days extends React.Component {
```
asci: could be replaced by stateless component

components/calendar/views/days.js
```diff
@@ -0,0 +1,305 @@
...
+
+    currentDate.setDate(currentDate.getDate() - numDaysOfPreviousMonth);
+
+    const options = [];
+    let counter = 0;
+
+    const limits = selectedDates.map(item => (item ? item.getTime() : null));
+    const selectedDatesAreEqual = (limits.length === 2) && (limits[0] === limits[1]);
+
+    while (counter < 42) {
```
asci: Answer to the Ultimate Question of Life, the Universe, and Everything?:D

components/calendar/calendar.js
mAiNiNfEcTiOn: how come?

components/calendar/views/days.js
mAiNiNfEcTiOn: Well I can add a description but 7 days * 6 rows = 42 options/buttons.

components/calendar/panels/days.js
mAiNiNfEcTiOn: Yes indeed... Will take care of that.

components/calendar/views/days.js
```diff
@@ -0,0 +1,305 @@
...
+      const extraClasses = [];
```
asci: maybe `mods`?

components/calendar/views/days.js
```diff
@@ -0,0 +1,305 @@
...
+    const isNextMonthDisabled = maxDate && (maxDate.getMonth() <= renderDate.getMonth());
+
+    return (
+      <nav className="ui-calendar-days__nav">
+        <button
+          className="ui-calendar-days__previous-month"
+          disabled={isPreviousMonthDisabled}
+          onClick={onNavPreviousMonth}
+          type="button"
+        >&lt;</button>
```
asci: here should be an icon, right?

components/calendar/views/days.js
mAiNiNfEcTiOn: Yeah I just didn't want it to be confused with the render's mods :)

components/calendar/views/days.js
mAiNiNfEcTiOn: Yes :) That's my next challenge 😄 ... Probably will expose via props what you can pass as nav buttons (next and previous).

themes/_default.yaml
```diff
@@ -23,8 +23,11 @@ generic:
     accent-darker: &accent-darker   '#0B4848'
     accent-dark: &accent-dark       '#177D7D'
     accent: &accent                 '#2A9595'
+    accent-border: &accent-border   '#E6BB00'
```
asci: this is yellow, not Lochinvar color, wich we use for an accent color in the default theme

mAiNiNfEcTiOn: Not what? :| the accent border was not there I added it lol.

Didn't even knew that we had any specific system to choose the colors 😮 .

Then you'll have to tell me how to select them ;-) (And I have to admit that the calendar colors look a little bit bad now 😄 )

components/calendar/calendar.js
asci: unless you're not modifying it in the render function it's ok. but somebody else could not notice it and add some
`mods.push('some-mod')` and it will change mods across all instances of this component. It is not a bug in your code, but potential weak point and I consider as a "better practice" not to put objects\arrays into default props object

mAiNiNfEcTiOn: Hmmm I see... good point. Will prevent that then :) 👍  thanks.

themes/_default.yaml
asci: http://paletton.com/palette.php?uid=73i0A0kmYv4gpRFk9G7qbq2qUf4 this was initial color system

mAiNiNfEcTiOn:
Now, let's go onto the icons and accessibility!

mAiNiNfEcTiOn:
Ok, now tackling the `componentWillReceiveProps` ... 😄

mAiNiNfEcTiOn:
Ok... now, support for setting the content of the next and previous buttons! 😄 

Also, I'll be adding the possibility to set the 'aria-label' sentence.

themes/_default.yaml
mAiNiNfEcTiOn: @asci are those colors better?

mAiNiNfEcTiOn:
Now, only missing the documentation (finally).

mAiNiNfEcTiOn:
@asci , @reaktivo , @fahad19, @AlexDudar, @EduardTrutsyk and @iwwwi , can you please review it now?

asci:
Are we going to reuse this calendar on our platfrom apart from home page? So far we have only one place where use calendar.

mAiNiNfEcTiOn:
@asci The CarTrawler widget will have a Search box which will use the calendar as well...

components/calendar/calendar.js
```diff
@@ -0,0 +1,314 @@
...
+ * @function processProps
+ * @param {Object} props Props to base the new state on.
+ * @param {Object} state (Existing) state to be based on for the existing values.
+ * @return {Object} New state to be set/used.
+ * @static
+ */
+function processProps(props) {
+  const { initialDates, maxDate, minDate, selectionType } = props;
+  const maxLimit = maxDate ? normalizeDate(new Date(maxDate), 23, 59, 59, 999) : null;
+  const renderDate = normalizeDate(((initialDates && initialDates.length && initialDates[0])
```
fahad19: does it have any side-effect? 

what happens to the original Date object, that is passed to this Calendar component from parent component?

mAiNiNfEcTiOn: @fahad19 it is mutated. But that's what happens when you call `.setHours()` or any of it's sister functions :)

fahad19: is there any specific reason for choosing the approach of `normalizeDate()` over `returnNewNormalizedDate()`?

mAiNiNfEcTiOn: @fahad19 check if it's better :)

mAiNiNfEcTiOn:
@AlexDudar thanks ;)

components/calendar/calendar.js
fahad19: I agree with you about what you mentioned about side-effects. The only problem we have here is that it is still referencing the parent Component's props.

I would always avoid props mutation of any kind in React. If I really need to transform the value in some way, and expect the Component to re-render on further changes, I would use `.setState()`, and scope it under the current Component only. Without affecting parent(s).

Similar issue was found in another PR few weeks ago. I think @asci wanted to created a Story out of it so all affected components are updated to avoid this issue.

PR #46, opened 2017-03-17
ESlint improvements

asci:
- add rule "prefer stateless components"
- add rule "ternary always requires a new line"
- fixes for these rules


mAiNiNfEcTiOn:
@asci you have some conflicts (probably caused by me, sorry 😞) but specially the prefer-stateless-function seem really nice! 👍

PR #61, opened 2017-04-06
Improvements for Modal component

EduardTrutsyk:
# What does this PR do:
* changed styles for Modal component
* updated theme
* changed settings for `autoprefixer`

3865f487d3b5e51cd53d9c967e46fa13

### DIFF
--- a/components/modal/modal.scss
+++ b/components/modal/modal.scss
@@ -39,7 +39,7 @@
   top: 50%;
   transform: translate(-50%, 10%);
   transition: all 200ms ease-in-out;
-  width: auto;
+  width: 100%;
   z-index: $tx-modal-z-index - 10;
 
   .ui-modal_fullscreen & {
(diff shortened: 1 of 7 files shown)


components/modal/modal.scss
```diff
@@ -80,41 +80,59 @@
...
 .ui-modal__close-button {
+  align-items: center;
   background: transparent;
   border: 0;
   color: $tx-modal-close-button-color;
   cursor: pointer;
-  line-height: 20px;
+  display: flex;
+  flex-direction: row-reverse;
+  font-size: 13px;
```
mAiNiNfEcTiOn: Would it be possible to use `rem` units? (in this case: `0.8125rem`)

Same for all other `px`... In terms of a11y and responsiveness is better: https://offroadcode.com/rem-calculator/

If you don't want, that's fine too... 👍

Issue #62, opened 2017-04-07
Can't pass custom modifications (eg. mods) to Modal component 


juliamaksimchik:
Hi everyone,

Currently I'm trying to restyle **Modal** component and I've discovered that there no ability to pass custom modification through props like for other components.

Could you please take a look at it? 

Thanks.

PR #63, opened 2017-04-07
Add posibility to pass mods through `props` in modal component

juliamaksimchik:
Fix for #62

### DIFF
--- a/components/modal/modal.js
+++ b/components/modal/modal.js
@@ -139,7 +139,7 @@ class Modal extends Component {
   }
 
   render() {
-    const { active, fullscreen, children, mods = [] } = this.props;
+    const { active, fullscreen, children, mods } = this.props;
     const { isOpen } = this.state;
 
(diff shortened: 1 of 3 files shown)


components/modal/modal.js
```diff
@@ -184,6 +186,7 @@ Modal.defaultProps = {
   closeOnOverlayClick: true,
   footer: null,
   fullscreen: false,
+  mods: [],
```
asci: possible reference issue, please put an array in the render function:
`const { active, fullscreen, children, mods = [] } = this.props;`

mAiNiNfEcTiOn: Well in this case it wouldn't happen, because of this: https://github.com/Travix-International/travix-ui-kit/pull/63/files#diff-aeb4723b4b18ec633e3c8bab2ea276e6R145

Nevertheless, to maintain consistency and prevent future problems, I agree with you @asci

PR #70, opened 2017-05-01
added package dependencies + missing list vars to fix the build

iwwwi:
## What does this PR do:
* adds the package dependency for `create-react-class` and `prop-types`
* adds variables that are used in the css but not defined in the `themes/_default.yaml`

## Where to start:
* diffs


package.json
```diff
@@ -58,10 +58,12 @@
     "babel-preset-travix": "^1.1.0",
     "babel-register": "^6.16.3",
     "commander": "^2.9.0",
+    "create-react-class": "^15.5.2",
     "css-loader": "^0.26.1",
     "extract-text-webpack-plugin": "^1.0.1",
     "node-sass": "^4.3.0",
     "postcss-loader": "^1.2.2",
+    "prop-types": "^15.5.8",
```
asci: sorry, why do we need this, if we have React `0.14`?

package.json
```diff
@@ -58,10 +58,12 @@
     "babel-preset-travix": "^1.1.0",
     "babel-register": "^6.16.3",
     "commander": "^2.9.0",
+    "create-react-class": "^15.5.2",
```
asci: would be better to use `classes`?

mAiNiNfEcTiOn: @asci that is a peerDependency which means that is used by one of our dependencies, not our code. It's to solve this issue: https://github.com/JedWatson/react-input-autosize/issues/82

PR #74, opened 2017-05-02
Update button.js

asci:
they are not implemented yet. this is misleading


components/button/button.js
```diff
@@ -101,7 +101,7 @@ Button.propTypes = {
   /**
    * Button's apperance.
    */
-  variation: PropTypes.oneOf(['default', 'ghost', 'social', 'rounded', 'borderless']),
+  variation: PropTypes.oneOf(['default']),
```
AlexDudar: we are going to add them later on right?

asci: yep, if there will be a need

Issue #77, opened 2017-05-05
Avoid mutating props in Components


fahad19:
As discussed here in February: https://github.com/Travix-International/travix-ui-kit/pull/16#discussion-diff-103207264R9

fahad19:
Closed via #78

PR #78, opened 2017-05-05
Button restProps and non-mutating mods

asci:
## What
- Button component `restProps` for data attributes for simple button fix
- Don't mutate props.mods (fix for #77 ): slice mods if provided or just use a new array

## Where
- Diff

## Tests
- 100% covereage
- added new snapshots
- updated old ones

### DIFF
--- a/components/autoComplete/autoCompleteItem.js
+++ b/components/autoComplete/autoCompleteItem.js
@@ -28,12 +28,12 @@ class AutoCompleteItem extends Component {
     const {
       children,
       dataAttrs = {},
-      mods = [],
       onClick,
       isActive,
       isTitle,
       ...otherProps
     } = this.props;
+    const mods = this.props.mods ? this.props.mods.slice() : [];
 
     isTitle && mods.push('title');
     isActive && mods.push('active');
--- a/components/button/button.js
+++ b/components/button/button.js
@@ -7,9 +7,19 @@ const { PropTypes } = React;
 /**
  * General Button component. Use when you need button or a link that looks like button
  */
-function Button({ children, mods = [], size, href, onClick, type, variation, disabled, dataAttrs = {} }) {
+function Button(props) {
+  const {
+    children,
+    size,
+    href,
+    onClick,
+    type,
+    variation,
+    disabled,
+    dataAttrs = {},
+  } = props;
   const restProps = getDataAttributes(dataAttrs);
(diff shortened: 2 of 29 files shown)


components/autoComplete/autoCompleteItem.js
```diff
@@ -28,12 +28,12 @@ class AutoCompleteItem extends Component {
     const {
       children,
       dataAttrs = {},
-      mods = [],
       onClick,
       isActive,
       isTitle,
       ...otherProps
     } = this.props;
+    const mods = this.props.mods ? this.props.mods.slice() : [];
```
reaktivo: I commented something similar in another PR, but seems to me that the following could be a good convention

    getClassNamesWithMods('ui-button', [
      isTitle && 'title',
      isActive && 'active',
    ]);

asci:
Any comments?

components/button/button.js
```diff
@@ -7,9 +7,19 @@ const { PropTypes } = React;
...
+    href,
+    onClick,
+    type,
+    variation,
+    disabled,
+    dataAttrs = {},
+  } = props;
   const restProps = getDataAttributes(dataAttrs);
-
+  const mods = props.mods ? props.mods.slice() : [];
```
reaktivo: I think it's a bad idea to have an arbitrary mods prop being passed here, specially considering there's no documentation for it and from what I see, should only affect internal behaviour (via the size and variation props)

asci: here is the docs for `mods`:
https://github.com/Travix-International/travix-ui-kit/pull/78/files/e6642dab2b515c3a613b7f0620ec86d6e3e41e4d#diff-22fbe4ecce9fc18f113d7f3df5fa1cc7R105 I guess the concept of mods is unclear: the mods property is for custom class selectors which developers need sometimes. And also we use it internally in components.

You can check how it works here https://travix-international.github.io/travix-ui-kit/#!/Button/1 there is an interactive playground.

The idea and the scope of this PR are to fix mutation of `mods` in all components, and not to change the way how they are work.

reaktivo: That was my point ` mods: PropTypes.arrayOf(PropTypes.string),` doesn't express at all how a developer would use it. It doesn't matter though, if we're moving to css-modules soon. And you're right about this is out of scope for this PR.

PR #87, opened 2017-05-10
Upgrade webpack 2

### LINKED ISSUE #84: Upgrade to Webpack 2
Otherwise applications using Webpack 2 themselves, cannot use `travix-ui-kit`.

As @iwwwi mentioned earlier, we basically need to revert this PR here: https://github.com/Travix-International/travix-ui-kit/pull/4

mAiNiNfEcTiOn:
# What does this PR do:
* Upgrades to Webpack 2 and removes `autoprefixer-loader` (deprecated) in favour of `postcss-loader`. Locks versions on minor.
* Changes one of the test cases of the `calendar.normal.spec.js` to not rely on snapshots (since dates change)
* Removes `enzyme-to-json` from the `calendar.normal.spec.js` as it is not used there anymore
* Updates the `webpack.config.js` to be Webpack 2 compatible

Should fix #84

  - Diffs
  - Do `npm install` since the `package.json` changed.
  - Run something like `rm dist/ui-bundle.* && npm run build`
  - Check if the bundles in the `dist/` are ok :)

All tests should be passing 😄


tests/unit/calendar/calendar.normal.spec.js
```diff
@@ -26,12 +25,8 @@ describe('Calendar (normal mode)', () => {
...
     it('should not mutate props', () => {
       const mods = ['test'];
-      const wrapper = shallow(
-        <Calendar
-          mods={mods}
-        />
-      );
-      expect(shallowToJson(wrapper)).toMatchSnapshot();
+
+      shallow(<Calendar mods={mods} />);
```
reaktivo: Instead of removing the shallow compare, why not test with a fixed date using the initialDates prop?

fahad19:
💖

tests/unit/calendar/calendar.normal.spec.js
mAiNiNfEcTiOn: According to the test, the idea is to just verify if the props are mutated. That does not require the shallowToJson

PR #94, opened 2017-05-18
Components props improvements

RumataDeEstor:
This PR adds possibility to pass data-attributes to Checkbox.


components/radioButton/radioButton.js
```diff
@@ -26,7 +27,11 @@ function RadioButton(props) {
   const className = getClassNamesWithMods('ui-radio', mods);
 
   return (
-    <div className={className} {...restProps}>
+    <div
+      className={className}
```
EduardTrutsyk: I think, the class name should be the last one.

components/checkbox/checkbox.js
```diff
@@ -8,16 +8,24 @@ function Checkbox(props) {
   const {
     checked,
     children,
+    dataAttrs = {},
     disabled,
     name,
     onChange,
+    ...otherProps
```
asci: what is `otherProps`?

RumataDeEstor: In other components, for example, Dropdown, Autocomplete, Input, we already have possibility to directly pass necessary props like `data-gtm-id` to component, which is quite convenient and laconically.

asci: why then you need `dataAttributes` (which were created exactly for `data-gtm-id` and other data attributes)?

RumataDeEstor: You are right. Moreover, I think that if we add possibility to pass `otherProps` to all components, we will get more flexibility, and, of course, we won't need ane `dataAttributes` in this case. So, I'll better remove `dataAttributes` from checkbox (because it was me who added them). But we can't remove `dataAttributes` from everywhere until we are sure the nobody uses it. It would be nice if we can make `dataAttributes` deprecated and fully switch to `otherProps` for flexibility.

EduardTrutsyk: I'm totally agree with you @RumataDeEstor

asci: In my opinion, it is better to remove `otherProps` because it provides too much power and leads to components misuse and unstable behavior. 

Btw, If you have `otherProps` you don't need ANY other props, so you can remove them as well :D

asci:
Unfortunately, I pressed Approve button too early and I don't see a way to unapprove. I personally don't support this changes, but I'm completely fine if the majority approve it.

components/radioButton/radioButton.js
```diff
@@ -26,7 +27,11 @@ function RadioButton(props) {
   const className = getClassNamesWithMods('ui-radio', mods);
 
   return (
-    <div className={className} {...restProps}>
+    <div
+      {...dataAttributes}
+      {...otherProps}
```
fahad19: is this unavoidable?

components/checkbox/checkbox.js
```diff
@@ -11,13 +11,19 @@ function Checkbox(props) {
...
   } = props;
+
   const mods = props.mods ? props.mods.slice() : [];
   disabled && mods.push('is-disabled');
   const className = getClassNamesWithMods('ui-checkbox', mods);
 
   return (
-    <label className={className} htmlFor={name}>
+    <label
+      {...otherProps}
```
fahad19: is it really needed?

PR #101, opened 2017-06-15
Added Badge component

maximuk:
# What does this PR do:
 - Added `Badge` component
 - `getClassNamesWithMods` function improvement
  - Diffs, specially the file `badge.js`.
Tests covering the component were added.
# Screenshots
selection_007
selection_008
selection_009


asci:
Would be nice to see some screenshots from LSG

components/badge/badge.js
```diff
@@ -0,0 +1,89 @@
...
+      'arrow-right': arrow && position === 'left',
+      'no-border': !border,
+    };
+
+    if (position && children) {
+      badgeMods[position] = true;
+    }
+
+    return (
+      <div {...otherProps} className={getClassNamesWithMods('ui-badge-badge', badgeMods)}>
```
asci: what suppose to be in `otherProps`?

tests/unit/badge/badge.renderBadge.spec.js
```diff
@@ -0,0 +1,42 @@
+import { shallow } from 'enzyme';
```
asci: why do you need 2 files?

components/badge/badge.js
maximuk: Some events like `onClick`, 'onMouseLeave' and so on

tests/unit/badge/badge.renderBadge.spec.js
maximuk: Because I test two different functions of the component

asci: will somebody call the `renderBadge` separately from `render`?

components/badge/badge.js
asci: Ok, I give up. Hopefully, nobody will misuse this prop

components/badge/badge.js
```diff
@@ -0,0 +1,89 @@
...
+   * The Badge's title
+   */
+  title: PropTypes.string,
+};
+
+Badge.defaultProps = {
+  arrow: false,
+  border: true,
+  children: null,
+  mods: [],
```
asci: potential object reference problem

tests/unit/badge/badge.renderBadge.spec.js
maximuk: no, but `render` can return just result from `renderBadge` in some cases, so I test them separately

components/_helpers.js
```diff
@@ -1,19 +1,21 @@
...
-    classes = classes
-      .concat(mods.filter(Boolean).map(mode => `${baseClass}${modSep}${mode}`));
-  } else {
-    Object.keys(mods).forEach((mode) => {
-      if (mods[mode]) {
-        classes.push(`${baseClass}${modSep}${mode}`);
-      }
-    });
-  }
+  modsSet.forEach((mods) => {
```
asci: from performance side it is not the best idea — it will affect the rendering of *all* components

tests/unit/badge/badge.renderBadge.spec.js
mAiNiNfEcTiOn: So, wouldn't that be the same file, but different cases?

components/_helpers.js
EduardTrutsyk: Seriously? micro optimization? better to write simple, more readable code, instead of try to micro optimize. I'm not sure if it will works more slowly then a lot of constructions like this: `if () {}`. 
BTW, at this moment all components has only one argument.
from my point of view, it's better way.
```
const className = getClassNamesWithMods('ui-badge-badge', this.props.mods, badgeMods);
```
*VS*
```
if (disabled) {
    mods.push(`disabled_true`);
}
```
In additional, you no need to create new `Array` in `render` method, because `getClassNamesWithMods` didn't change mobs from props, just return `string`

tests/unit/badge/badge.renderBadge.spec.js
maximuk: I think tests in separated files are more clear

mAiNiNfEcTiOn: I also thought that way, but concentrating all in one file, mostly grouped by rendering differences and/or interactions, seems waaay more reasonable :)

EduardTrutsyk: better to have small files, because:

1. all test in one file related to one function (single responsibility principle + separation of concerns)
2. in this case we have separate file with `snapshots`
3. reduce conflicts
4. easy to find failed test
5. same approach like we have on core project(rwd)
from my point of view we have only pros without cons ;)

mAiNiNfEcTiOn: @EduardTrutsyk I also thought like that before - you can see from the amount of projects we have in rwd - but that will change in the future specially for rwd :)

The reasons why a one file is better, is related to what we're testing.

We're testing how our component behaves when rendering and/or interacting with it (simulating click events, etc).

This is definitely a good way to test integration between apps, and test if the app's functionality is broken.

I don't see that in our approach on rwd. We test really well our functions, but that only gives us a good coverage for each unit, not a proper sign that we're breaking the intended functionality or changing it's behaviour :)

components/badge/badge.js
```diff
@@ -0,0 +1,87 @@
...
+  Component,
+  PropTypes,
+} from 'react';
+
+import { getClassNamesWithMods } from '../_helpers';
+
+/**
+ * Badge component
+ */
+class Badge extends Component {
```
asci: Why is it a stateful component? I don't see any state management here

tests/unit/badge/badge.renderBadge.spec.js
asci: 1. If we talking about single responsibility and unit testing then you need to mock any call from `render` to `renderBadge`. Another point: would be nice to have files per public (part of the contract) function, not just a private helpers (function called only inside the file). You can easily replace that function call with a local variable in `render` function and reduce the number of files and function calls.
2. You have 2 the same snapshots
3. How would you reduce conflicts if your snapshots produced by one part of the code? I'd say you will increase it;)
4. Yep, you *just* need to select **correct** file
5. But we don't have it here, would be nice to follow pattern of current package?

components/_helpers.js
asci: On one side, this function will be called thousands of times by thousands of components on a page and it is important to make it as performant as possible (yes, you can call it "micro-optimization", but it is an optimization for frequently called function)
On the other side, I agree with you about readability.

tests/unit/badge/badge.renderBadge.spec.js
mAiNiNfEcTiOn: Not to brag about this, but the `Calendar`'s spec is a good example when it comes to use a single file to test something.

However if you look at `tests/unit/calendar/` you will see 2 files (_.normal._ and _.range._). _Why?_ you may ask... Well, I decided to split them on the 'type' of calendar, since it changes the expectations and/or main behavior/rendering process. In that case it does simplify the reading and maintainability.
Could they be in the same file? Yes... Should they? _Maybe_, but I did a 'virtual' separation.

So, @maximuk you're free to do a separation on those functions, though it is my strong belief that you should only split them in case you cannot _touch_ the functions via normal render :)

Last but not the least, if you're wondering why the .specs. do not use the `matchSnapshot()` methods, it is mainly due to the fact that there are `Date`s involved in the rendering... Not that easy to maintain as the snapshots would keep changing :) Usually it fits, this is a special case 😄 

Hope I was clear.

PR #104, opened 2017-06-21
Handle building intermediate directories when outputing files

reaktivo:
If we don't have a dist directory, the build fails, this PR removes the need to have any intermediate directories when building.


package.json
```diff
@@ -69,6 +70,7 @@
     "eslint-plugin-react": "^6.10.2",
     "eslint": "^3.14.0",
     "extract-text-webpack-plugin": "~2.1.0",
+    "fs-extra": "^3.0.1",
```
asci: why is it repeated?

reaktivo: Don't know why it's not updating the PR, but I've updated the branch already

package.json
```diff
@@ -45,6 +45,7 @@
     }
   ],
   "dependencies": {
+    "fs-extra": "^3.0.1",
```
asci: 😞

package.json
reaktivo: @asci Fixed now.

asci: Why not `mkdirp`?

mAiNiNfEcTiOn: Well, it's a dev dependency, nobody dies. But if you really want to be cool, just do:

```bash
mkdir dist
touch dist/.gitkeep
```

And add to the .gitignore to:

```
!dist/.gitkeep
```

reaktivo:
Fixing tests!

package.json
reaktivo: @mAiNiNfEcTiOn This would still break when specifying a different dist directory.

@asci I prefer `outputFile` from `fs-extra`, since it perfectly matches the function signature from `saveThemeScssFile`.

mAiNiNfEcTiOn: @reaktivo specifying a different `dist` folder? Hmmm, this is the folder used for the 'deliverable' of the package. It won't be built anymore on rwd :) Only copied ;) :)

reaktivo: @mAiNiNfEcTiOn I thought for a moment that this was user configurable, oh well.

Issue #73, opened 2017-05-02
Button variations is not working


ibsukru:
https://travix-international.github.io/travix-ui-kit/ Have tried ghost, rounded .. but no success

asci:
They are not implemented yet (:

ibsukru:
Oki, Better to update documentation, we were pointing our integration with ui kit tasks based on this features, it is misleading.

yurist38:
@ibsukru should we close this issue now or it's still relevant?

PR #106, opened 2017-06-22
adds ghost button

froskie:
Adds the ghost button variation to UI-KIT

image

image

This PR solves the _ghost_ rendering issue on #73


components/button/button.scss
```diff
@@ -46,6 +46,23 @@
   }
 }
 
+.ui-button_variation_ghost {
```
asci: any disabled case?

froskie: Not according to UI (last row):

image

PR #111, opened 2017-06-27
Adds the SlidingPanel component

mAiNiNfEcTiOn:
# What does this PR do:

* Adds the `SlidingPanel` component
* Adds the `SlidingPanel` SCSS file
* Adds the `SlidingPanel` Markdown file
* Adds the `SlidingPanel` unit tests

screen shot 2017-06-27 at 14 34 32

Note: Closes issue #110 

  - Diffs

Should be making 100% coverage.

### DIFF
--- a/components/slidingPanel/slidingPanel.js
+++ b/components/slidingPanel/slidingPanel.js
@@ -51,6 +51,10 @@ export default class SlidingPanel extends Component {
     open: PropTypes.bool,
   }
 
+  static defaultProps = {
+    closeOnOverlay: true,
+  }
+
   componentWillReceiveProps(newProps) {
     if (newProps.open !== this.state.isOpen) {
       if (newProps.open) {
--- a/components/slidingPanel/slidingPanel.js
+++ b/components/slidingPanel/slidingPanel.js
@@ -2,8 +2,6 @@ import React, { Component, PropTypes } from 'react';
 import { findDOMNode } from 'react-dom';
 import { getClassNamesWithMods, getDataAttributes } from '../_helpers';
 
-import Global from '../global/global';
-
 export default class SlidingPanel extends Component {
   constructor(props) {
     super(props);
@@ -75,7 +73,8 @@ export default class SlidingPanel extends Component {
 
     this.panel.addEventListener('transitionend', this.handleTransitionEnd);
 
-    const closeButton = this.panel.querySelector('[rel="close"]');
+    const rootNode = findDOMNode(this);
+    const closeButton = rootNode.querySelector('[rel="close"]');
     if (closeButton) {
       closeButton.addEventListener('click', this.handleClose);
     }
@@ -158,17 +157,15 @@ export default class SlidingPanel extends Component {
     const overlayClassName = getClassNamesWithMods(overlayClass, overlayMods);
 
     return (
-      <Global>
-        <div className={overlayClassName} onClick={this.handleClickOverlay}>
-          <div
-            className={panelClassName}
-            ref={e => (this.panel = e)}
-            {...getDataAttributes(dataAttrs)}
-          >
-            {children}
-          </div>
+      <div className={overlayClassName} onClick={this.handleClickOverlay}>
+        <div
+          className={panelClassName}
+          ref={e => (this.panel = e)}
+          {...getDataAttributes(dataAttrs)}
+        >
+          {children}
         </div>
-      </Global>
+      </div>
     );
   }
 }
--- a/components/slidingPanel/slidingPanel.js
+++ b/components/slidingPanel/slidingPanel.js
@@ -2,6 +2,8 @@ import React, { Component, PropTypes } from 'react';
 import { findDOMNode } from 'react-dom';
 import { getClassNamesWithMods, getDataAttributes } from '../_helpers';
 
+import Global from '../global/global';
+
 export default class SlidingPanel extends Component {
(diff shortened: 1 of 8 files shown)


components/slidingPanel/slidingPanel.js
```diff
@@ -0,0 +1,171 @@
...
+  static propTypes = {
+    children: PropTypes.oneOfType([
+      PropTypes.arrayOf(PropTypes.node),
+      PropTypes.node,
+    ]).isRequired,
+
+    /**
+     * When true, if the user clicks on the overaly, closes the panel.
+     */
+    closeOnOverlay: PropTypes.bool,
```
reaktivo: For consistency's sake, can we use the same prop name as in the Modal? `closeOnOverlayClick`

components/slidingPanel/slidingPanel.js
```diff
@@ -0,0 +1,171 @@
...
+
+    /**
+     * When defined, this function is triggered when the panel is opening.
+     */
+    onOpen: PropTypes.func,
+
+    /**
+     * Defines if the panel is open.
+     */
+    open: PropTypes.bool,
```
reaktivo: Same as previous one, can we use `active` ? I don't particularly love the name, but it's better if it's consistent.

components/slidingPanel/slidingPanel.js
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  render() {
+    const {
+      dataAttrs,
+      children,
+    } = this.props;
+
+    const overlayMods = [];
+    const panelMods = this.props.mods ? this.props.mods.slice() : [];
```
reaktivo: I've been trying to move away from mutating array's in some of the code introduced in `ui-kit`, what do you think about handling it the following way:

```
const panelClassName = getClassNamesWithMods(panelClassName, [
  this.state.isOpen && 'open',
  ...(this.props.mods || []),
]);
const overlayClassName = getClassNamesWithMods(overlayClass, [
  this.state.hideOverlay && 'hide',
]);
```

themes/_default.yaml
```diff
@@ -372,6 +373,26 @@ price:
       integers: 51px
     height:
       underline: 8px
+sliding-panel:
+  overlay:
+    background-color: "var(--tx-generic-color-primary-dark-rgb, .65)"
```
asci: we don't support default values for CSS variables

components/slidingPanel/slidingPanel.js
asci: now `getClassNamesWithMods` supports variable arity https://github.com/Travix-International/travix-ui-kit/blob/master/components/_helpers.js#L3 you can use rest arguments as mods:
```js
getClassNamesWithMods('base', myMods, mods);
```

AlleeX:
Guys, what about stacking context?

components/slidingPanel/slidingPanel.js
mAiNiNfEcTiOn: 🙄  sure...

components/slidingPanel/slidingPanel.js
mAiNiNfEcTiOn: `active`?

What does `active` means ? In which activity? AFAIK the panel is either open or closed, not active.

components/slidingPanel/slidingPanel.js
mAiNiNfEcTiOn: @reaktivo 
? What? Can you explain to me where am I mutating the arrays?

@asci 
Regarding variable arity, what would it be solving here? I'm adding to the passed `mods` (sliced) the mods that I need in case of conditional. So I would still need an array ...

\#microOptimizationsFTW

themes/_default.yaml
mAiNiNfEcTiOn: Actually that's a bug... 😄  good catch.

components/slidingPanel/slidingPanel.js
reaktivo: I agree that the name is bad, as I mentioned before, but I think it's way worst if we have components accept different prop naming for the same use cases. If it were that easy, I would rename the `active` prop in Modal, but it's already in use.

components/slidingPanel/slidingPanel.js
reaktivo: It's clear that it's exactly the same end result here. It's not about micro-optimizations, it's about encouraging developers to use arrays in a non mutating way, it's really easy to introduce hard to catch errors this way, (they've been introduced this way before), if someone forgets to slice or concat.

`Array.prototype.push` is naturally mutating.

mAiNiNfEcTiOn: @reaktivo mutating my own arrays. Can you explain where it would cause an issue or improve performance?

That is all interesting but the fact is the code is far more readable and predictable than trust in _multi-arity_ of a function.

mAiNiNfEcTiOn:
@AlleeX I will - for now - attempt to get away with it without using the `Global` component.

If for some reason we decide that we need it, I'll refactor it to use it, just like the Modal ;-)

Thanks for your input!

components/slidingPanel/slidingPanel.js
asci: IMO, easier to read:
```
const panelMods = [];

getClassNamesWithMods('base', panelMods, mods);
```
You're right, it is not about performance

PR #112, opened 2017-06-27
Adds the rating component

### LINKED ISSUE #113: Rating

froskie:
## What this PR do
- Adds the SlidingPanel component
- Adds the SlidingPanel SCSS file
- Adds the SlidingPanel Markdown file
- Adds the SlidingPanel unit tests

image

Closes #113 

- Diffs

- Should be making 100% coverage.

### DIFF
--- a/components/rating/rating.scss
+++ b/components/rating/rating.scss
@@ -1,0 +1,12 @@
+.ui-rating {
+  color: var(--tx-generic-color-secondary);
+  position: relative;
+  display: inline-block;
+}
+.ui-rating-value {
+  color: var(--tx-generic-color-accent);
+  position: absolute;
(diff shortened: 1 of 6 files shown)


components/rating/rating.scss
```diff
@@ -0,0 +1,12 @@
+.ui-rating {
```
asci: could be just `.css` (we'll slightly move off of SASS)

mAiNiNfEcTiOn: It could yes... But for consistency purposes and until we change to .css on all, I'd rather let this as it is.

froskie: Got it. Just followed the pattern on the other folders. But thanks 👍

Issue #109, opened 2017-06-26
Modal component is still rendered (although transparent) after closed by clicking on overlay


reaktivo:
When I create a Modal component in the following way:

```js
<Button size="s" onClick={() => setState({ isOpenNormalModal: true })}>Open normal modal</Button>
<Modal active={state.isOpenNormalModal}>
  Modal Content
</Modal>
```

I would expect that clicking on the background overlay would hide the modal and allow the page to be clicked/scrolled as normal. If the `onClose` prop is not set to a function that sets the `active` prop, the background overlay becomes transparent, but still clickable, and prevents normal interaction with the page.

I tried a few possible fixes, but those introduced other buggy behaviour.

![buggy_modal](https://user-images.githubusercontent.com/13637/27542400-70a8ed2e-5a87-11e7-942e-62a1b6476853.gif)

asci:
would
```css
.hidden .background {
pointer-event: none;
}
```

fix this?

reaktivo:
Nope, the problem is that the Modal is still being rendered and while rendered it prevents the page from being scrolled.

PR #115, opened 2017-06-28
#109 modal bug

froskie:
## What this PR do

- Fix for #109 

- Diffs

- Should be making 100% coverage.

### DIFF
--- a/components/modal/modal.js
+++ b/components/modal/modal.js
@@ -40,7 +40,8 @@ class Modal extends Component {
 
   open() {
     global.window.requestAnimationFrame(() => {
-      setTimeout(() => this.setState({ isActive: true, isOpen: true }), 0);
+      this.setState({ isActive: true });
+      setTimeout(() => this.setState({ isOpen: true }), 300);
     });
   }
 
@@ -152,7 +153,7 @@ class Modal extends Component {
       mods.push('active');
     }
 
-    if (!isOpen) {
+    if (!isOpen && !isActive) {
       return null;
     }
 
--- a/components/modal/modal.js
+++ b/components/modal/modal.js
@@ -8,10 +8,15 @@ import { getClassNamesWithMods } from '../_helpers';
  * Modal component
  */
 class Modal extends Component {
-  state = {
-    isActive: false,
-    isOpen: false,
-  };
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      isActive: props.active || false,
+      isOpen: false,
+    };
+  }
 
   componentDidMount() {
     if (this.props.closeOnEsc) {
--- a/components/modal/modal.js
+++ b/components/modal/modal.js
@@ -9,6 +9,7 @@ import { getClassNamesWithMods } from '../_helpers';
  */
 class Modal extends Component {
   state = {
+    isActive: false,
     isOpen: false,
   };
 
@@ -39,7 +40,7 @@ class Modal extends Component {
 
   open() {
     global.window.requestAnimationFrame(() => {
-      setTimeout(() => this.setState({ isOpen: true }), 0);
+      setTimeout(() => this.setState({ isActive: true, isOpen: true }), 0);
     });
   }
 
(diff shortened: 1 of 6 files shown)


tests/unit/modal/__snapshots__/modal.handleOverlayClick.spec.js.snap
```diff
@@ -1,22 +1,3 @@
...
-  >
-    <section
-      className="ui-modal__content"
-    >
-      Modal Content
-    </section>
-  </div>
-</Global>
-`;
+exports[`Modal: handleOverlayClick should call "onOverlayClick" prop after click on overlay 1`] = `null`;
```
asci: why `null`?

components/modal/modal.js
```diff
@@ -142,15 +145,15 @@ class Modal extends Component {
   }
 
   render() {
-    const { active, fullscreen, children } = this.props;
```
asci: How to open modal by default?

components/modal/modal.js
```diff
@@ -9,6 +9,7 @@ import { getClassNamesWithMods } from '../_helpers';
  */
 class Modal extends Component {
   state = {
+    isActive: false,
```
mAiNiNfEcTiOn: Have you considered adding a constructor here to set the state based on props initially:

```js
constructor(props) {
  super(props);

  this.state = {
    isActive: props.active || false,
  };
}
```

?

tests/unit/modal/__snapshots__/modal.handleOverlayClick.spec.js.snap
mAiNiNfEcTiOn: I think it's related to the same issue as above. Because it's no longer listening to the `active` prop to determine if it's active or not, this condition is met:

https://github.com/Travix-International/travix-ui-kit/pull/115/files#diff-aeb4723b4b18ec633e3c8bab2ea276e6R156

Because it is relying on the `active` prop...
https://github.com/Travix-International/travix-ui-kit/blob/fa3e4c9b54163ea12175efac8d3d3f7a82c8b55a/tests/unit/modal/modal.handleOverlayClick.spec.js#L12

(That's why this `matchSnapshot()` it's dangerous. If the developer updates the snapshot without noticing what changed, it is problematic.)

Good catch @asci !

components/modal/modal.js
froskie: The method `componentDidMount` check for that property and call the `open` method.

components/modal/modal.js
froskie: Although it's fine, again, the method `componentDidMount` check for that property and call the `open` method.

tests/unit/modal/__snapshots__/modal.handleOverlayClick.spec.js.snap
asci: >>If the developer updates the snapshot without noticing what changed
agree, it is dangerous, but that's why we have pull requests :)
On the other hand developer can change expected values manually to pass the test (I saw it couple of times) :D

components/modal/modal.js
```diff
@@ -8,9 +8,15 @@ import { getClassNamesWithMods } from '../_helpers';
...
 class Modal extends Component {
-  state = {
-    isOpen: false,
-  };
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      isActive: props.active || false,
```
mAiNiNfEcTiOn: Since you added this (nice one 👍 ), do you still need this one: https://github.com/Travix-International/travix-ui-kit/pull/115/files#diff-aeb4723b4b18ec633e3c8bab2ea276e6R25 ?

reaktivo:
Although this is better than the previous implementation. Maybe the ideal solution would be to follow other react Modal implementations, and only allow the `active` and `inactive` state to be handled from outside (see `onRequestClose` from [react-modal](https://github.com/reactjs/react-modal)).

This is what happens if you don't reset the `active` prop:

modal without onclose

PR #116, opened 2017-06-29
Change SlidingPanel to use the refs

mAiNiNfEcTiOn:
# What does this PR do:
* Adds usage of refs to remove usage of `findDOMNode`.

  - Diffs

Adds `renderTree.unmount()` to pass via `componentWillUnmount()`.


froskie:
There's a duplicate call to `jest.runAllTimers()` here:

https://github.com/Travix-International/travix-ui-kit/blob/1154957121e45a8fc5ac03a8f1b8b310085aa3a0/tests/unit/slidingPanel/slidingPanel.spec.js#L44

mAiNiNfEcTiOn:
@froskie good catch ... fixing 👍

PR #125, opened 2017-07-11
Collapse component adjustments

RumataDeEstor:
This PR provides the possibility to customize the position of the icon which is displayed on every collapse item (either left or right, left by default).


RumataDeEstor:
screenshot 2017-07-11 17-38-53

PR #126, opened 2017-07-11
Carousel component

froskie:
## What this PR do
- Adds the carousel component
image

- Diffs

- 100% on the carousel folder


AlleeX:
linting

components/carousel/carousel.js
```diff
@@ -0,0 +1,177 @@
...
+      next = 0;
+    }
+
+    this.setNextItem(next);
+  }
+
+  handleSwipePrev() {
+    let next = this.state.currentItem - 1;
+    if (next < 0) {
+      return;
```
asci: why not support indefinite loop?

froskie: Because when swiping, it'd be _unnatural_ to swipe and throw all the way to the other side. It's in the roadmap to do the trick about adding the last before the first and first after last in order to create that.

components/carousel/carousel.scss
```diff
@@ -0,0 +1,86 @@
+.ui-carousel {
```
asci: could be just CSS?

components/carousel/carouselMarkers.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default function CarouselMarkers({ images, current, onClick }) {
+  return (
+    <div className="ui-carousel-markers">
+      {images.map((src, i) => (i === current ? <i key={i} /> : <button data-index={i} key={i} onClick={onClick} />))}
```
asci: `<i key={i} />` this is not sematically right
https://www.w3.org/International/questions/qa-b-and-i-tags they it should represent *italic* content

components/carousel/carouselPage.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default function CarouselPage({ images, current }) {
+  return (
+    <div className="ui-carousel-page">
+      {current + 1} / <b>{images.length}</b>
```
asci: consider to use `span`

components/carousel/carousel.scss
```diff
@@ -0,0 +1,86 @@
...
+  transition: var(--tx-carousel-track-transition);
+}
+
+.ui-carousel-item {
+  flex: 1 0 100%;
+  position: relative;
+  min-height: var(--tx-carousel-item-min-height);
+}
+
+.ui-carousel-item-image {
```
AlleeX: what about rule "styles sorting" for ui kit? Should we do it by alphabetical or type groups?
@mAiNiNfEcTiOn  @asci

themes/_default.yaml
```diff
@@ -401,3 +401,29 @@ spinner:
...
     dark: "var(--tx-generic-color-active)"
     darker: "var(--tx-generic-color-primary-lighter)"
+carousel:
+  track:
+    transition: "transform 0.5s cubic-bezier(0.2, 1, 0.3, 1)"
+  item:
+    min-height: "400px"
+  button:
+    font-size: "24px"
+    background-color: "var(--tx-generic-color-blank)"
```
asci: why not
```
background-color: *blank 
```
?

utils/swipe.js
```diff
@@ -0,0 +1,97 @@
...
+    this.handleEnd = this.handleEnd.bind(this);
+
+    this.$elm = typeof elm === 'string' ? document.querySelector(elm) : elm;
+    this.hasTouch = 'ontouchstart' in window;
+    this.bindEvents();
+  }
+
+  bindEvents() {
+    if (this.hasTouch) {
+      this.$elm.addEventListener('touchstart', this.handleStart, false);
```
asci: and where do you remove listeners?

components/carousel/carousel.scss
asci: I believe this shouldn't be something we can talk about, I'm fine with non-sorted alphabetically, but since we have it in many other repos, would be nice to be consistent. But I also believe that this could be automated by something like this: https://github.com/morishitter/stylefmt

utils/swipe.js
froskie: I've just added the code below. It'll be easier when changed from class to component.

components/carousel/carouselItem.js
```diff
@@ -0,0 +1,50 @@
...
+    }
+    this.setState(() => ({ hasLoaded: true, imageSrc: src }));
+  }
+
+  render() {
+    const { imageSrc, hasLoaded } = this.state;
+    const imageStyles = {};
+
+    if (imageSrc && hasLoaded) {
+      imageStyles.backgroundImage = `url('${imageSrc}')`;
```
reaktivo: Looks like you don't need to keep track both of `imageSrc` and `hasLoaded` using `props.src` and `state.hasLoaded` is enough.

components/carousel/carousel.css
```diff
@@ -0,0 +1,83 @@
+.ui-carousel {
+  overflow: hidden;
+  position: relative;
+}
+
+.ui-carousel-item {
```
reaktivo: I think most styles are using the BEM style of defining blocks, basically, blocks like this should be defined in this style `.ui-carousel__item`, `.ui-carousel__item-item`, `.ui-carousel__markers`, etc.

utils/swipe.js
```diff
@@ -0,0 +1,109 @@
...
+      this.$elm.removeEventListener('touchmove', this.handleMove, false);
+      this.$elm.removeEventListener('touchend', this.handleEnd, false);
+    } else {
+      this.$elm.removeEventListener('mousedown', this.handleStart, false);
+      this.$elm.removeEventListener('mousemove', this.handleMove, false);
+      this.$elm.removeEventListener('mouseup', this.handleEnd, false);
+    }
+  }
+
+  onLeft(callback) {
```
reaktivo: This style of API is a bit strange, since it doesn't allow you to redefine a callback handler, since it redefines itself, which is not obvious, can you consider the following change?

```
onLeft(callback) {
  this.onLeftHandler = callback;
  return this;
}
```

froskie: I didn't understand. Why redefine a callback?

components/carousel/carousel.css
froskie: Although I understand and also code like that, @reaktivo, from other source codes I've read, this isn't really being enforced. So I'd like to know if we're going to do so. Not only that, the helper that generates the mods class names doesn't apply mods as '--'.

Any thoughts on that, @mAiNiNfEcTiOn @asci ?

components/carousel/carousel.js
```diff
@@ -0,0 +1,181 @@
...
+        <div className="ui-carousel-track" ref={(c) => { this.carouselTracker = c; }} style={trackerStyles}>
+          {this.state.images.map((src, i) => <CarouselItem key={i} load={this.shouldLoad(i)} src={src} />)}
+        </div>
+        {this.state.images.length > 1 && this.renderNavigation()}
+      </div>
+    );
+  }
+}
+
+Carousel.propTypes = {
```
froskie: Added.

components/carousel/carousel.css
asci: I agree with @reaktivo that naming could be done better, but for me, it is not critical since components are quite isolated

utils/swipe.js
froskie: Settled with @reaktivo on a direct talk.

components/carousel/carousel.md
```diff
@@ -0,0 +1,36 @@
+Carousel:
+
+    <div>
+      <Carousel
+        images={[
+          'http://lorempixel.com/600/400/city',
```
EduardTrutsyk: is it possible to use custom tag with text or something like that? or can I add custom attributes to img tag?

froskie: Not on this version, @EduardTrutsyk. There's a roadmap for a _generic carousel_. You'd input the children elements.

mAiNiNfEcTiOn: Yeah something like:

```html
<Carousel ...>
  <CarouselItem>
    <div className="your-div">
      Your content
    </div>
  </CarouselItem>
</Carousel>
```

AlleeX: We can create issue for it and do it in the next step

tests/unit/carousel/carousel.spec.js
```diff
@@ -0,0 +1,193 @@
...
+
+describe('Carousel', () => {
+  describe('#render()', () => {
+    it('render with default props and mods', () => {
+      const images = [
+        'http://lorempixel.com/600/400/city',
+        'http://lorempixel.com/600/400/sports',
+        'http://lorempixel.com/600/400/people',
+      ];
+      const renderTree = mount(
```
EduardTrutsyk: why are you using `mount` here? for real unit tests better to use `shallow`

PR #133, opened 2017-07-26
Message box

iwwwi:
# What does this PR do:

* creates message box component - wrapper for messages/info/error
* adds styles
* adds unit test

* diffs

# Screenshots
selection_013


components/messageBox/messageBox.js
```diff
@@ -0,0 +1,50 @@
+import React, { PropTypes } from 'react';
+import { getClassNamesWithMods, getDataAttributes } from '../_helpers';
+
+const MessageBox = (props) => {
+  const { children, dataAttrs, isError } = props;
```
EduardTrutsyk: is it possible to add `title`?
screenshot at 09-43-41

components/messageBox/messageBox.js
```diff
@@ -0,0 +1,50 @@
...
+  const mods = props.mods ? props.mods.slice() : [];
```
EduardTrutsyk: in my opinion, it looks much better: 
`const mods = getClassNamesWithMods('ui-messageBox', props.mods, { error: isError });`
instead of 
```
if (isError) {
   mods.push(`error`);
}
```

components/messageBox/messageBox.js
```diff
@@ -0,0 +1,50 @@
...
+
+  /**
+   * You can provide set of custom modifications.
+   */
+  mods: PropTypes.arrayOf(PropTypes.string),
+
+  /**
+   * Set to true if this represents an error message.
+   */
+  isError: PropTypes.bool,
```
EduardTrutsyk: what about `success`(green), `info`(blue), `warning`(yellow)?
screenshot at 09-54-10

components/messageBox/messageBox.js
```diff
@@ -1,19 +1,30 @@
...
-  const { children, dataAttrs, isError } = props;
-  const mods = props.mods ? props.mods.slice() : [];
+const MessageBox = ({ children, dataAttrs, icon, mods, title, type }) => {
+  const className = getClassNamesWithMods('ui-messageBox', [...mods, type]);
 
-  if (isError) {
-    mods.push(`error`);
-  }
+  const logo = icon && (
+    <div className="ui-messageBox__content__icon">
```
yurist38: I'm pretty sure this naming is not fit to the BEM convension. It doesn't allow to use `__` separator twice in one class name. Should be smth like `ui-messageBox__content-icon`. Please fix it here and in all other places...

mAiNiNfEcTiOn: @maximuk can you please fix this as well?

components/messageBox/messageBox.js
```diff
@@ -0,0 +1,77 @@
+import React, { PropTypes } from 'react';
```
fahad19: please import PropTypes from `prop-types` package. otherwise we keep adding more to our debt when upgrading to React v15+.

PR #137, opened 2017-08-08
Add header to SlidingPanel component

yurist38:
## WHAT DOES THIS PR DO: ##

* Added `slidingPanelHeader` component (for now internally, inside `slidingPanel`). 
* Added optional prop `title` to `slidingPanel` component. If it's defined we do show header.
* `slidingPanel` component is changed to support multiple close buttons.

## SCREENSHOT: ##

slidingpanelpreview

* Diffs

## UNIT TESTS: ##

* Added and adjusted to achieve 100% coverage

### DIFF
--- a/components/slidingPanel/slidingPanel.js
+++ b/components/slidingPanel/slidingPanel.js
@@ -1,4 +1,6 @@
-import React, { Component, PropTypes } from 'react';
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import SlidingPanelHeader from './slidingPanelHeader';
 import { getClassNamesWithMods, getDataAttributes } from '../_helpers';
 
 export default class SlidingPanel extends Component {
@@ -48,6 +50,11 @@ export default class SlidingPanel extends Component {
      * Defines if the panel is open.
      */
     active: PropTypes.bool,
+
+    /**
+     * Defines title for header. Optional. If it's defined header will be shown.
+     */
+    title: PropTypes.string,
   }
 
   static defaultProps = {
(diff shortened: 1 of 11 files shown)


components/slidingPanel/slidingPanel.js
```diff
@@ -1,4 +1,6 @@
-import React, { Component, PropTypes } from 'react';
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
```
asci: 👍

components/slidingPanel/slidingPanelHeader.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const SlidingPanelHeader = ({ title }) => {
+  if (!title) {
+    return <noscript />;
```
asci: afaik could be `null`

tests/unit/slidingPanel/slidingPanel.spec.js
```diff
@@ -8,19 +8,20 @@ describe('SlidingPanel', () => {
   describe('#render()', () => {
     it('render with default props and mods provided', () => {
       const renderTree = mount(
-        <SlidingPanel mods={['test']}>Test</SlidingPanel>
+        <SlidingPanel mods={['test']} title="Test Title">Test</SlidingPanel>
```
asci: this is not `default props`, please add new test

yurist38: Ah, okay. Got it. Will do now

components/slidingPanel/slidingPanelHeader.js
yurist38: It was `null` in my first implementation. But then I stuck with testing, it causes [this](https://github.com/facebook/jest/issues/642) error. Seems like unless we migrated to React 15 we still need to use `noscript`. By the way, I've searched other components and found many places where we also use `noscript`. Once we migrated I think could be nice to get rid of them.

mAiNiNfEcTiOn: @asci https://github.com/facebook/react/issues/5355

asci: I forget every time:D When we'll migrate?

PR #138, opened 2017-08-08
Google map component

RumataDeEstor:
This PR adds GoogleMap Component which is powered by Google API script added to main page. 
screenshot 2017-08-08 15-34-33


components/googleMap/googleMap.js
```diff
@@ -0,0 +1,306 @@
+import React, { Component } from 'react';
+import { PropTypes } from 'prop-types';
```
fahad19: I wasn't aware PropTypes can be imported like this (destructured) from `prop-types` package too.

RumataDeEstor: *unified*

fahad19: both seem to work 👍 

tried out here: https://codesandbox.io/s/KrmEjrmpz

would be nice if the project sticks to one way of using it though across all the components.

components/googleMap/googleMap.js
```diff
@@ -0,0 +1,306 @@
...
+      this.latlng = position;
+      this.info = info;
+      this.index = index;
+      this.setMap(map);
+    }
+
+    CustomMarker.prototype = new google.maps.OverlayView();
+
+    CustomMarker.prototype.draw = function draw() {
+      let div = this.div;
```
asci: Is it possible to move this outside the `GoogleMap` component class?  having different `this` is really confusing

components/googleMap/googleMap.js
```diff
@@ -0,0 +1,306 @@
...
+    function CustomMarker(pointData, map, index) {
+      const { position, info } = pointData;
+
+      this.latlng = position;
+      this.info = info;
+      this.index = index;
+      this.setMap(map);
+    }
+
+    CustomMarker.prototype = new google.maps.OverlayView();
```
asci: any checks that `google` is defined?

components/googleMap/googleMap.js
```diff
@@ -0,0 +1,306 @@
...
+      let div = this.div;
+      if (!div) {
+        div = document.createElement('DIV');
+        this.div = div;
+        div.style.position = 'absolute';
+        div.style.paddingLeft = '0px';
+        div.style.cursor = 'pointer';
+        div.style.width = '10px';
+        div.style.height = '10px';
+        div.classList.add('ui-google-map__custom-marker');
```
asci: is it possible to move all `div.style` under this class rule in CSS?

components/googleMap/googleMap.js
RumataDeEstor: @fahad19 yep, I fixed it. Let's keep using as a default export :)
but interesting to know :)

components/googleMap/googleMap.js
```diff
@@ -0,0 +1,306 @@
...
+    });
+  }
+
+  /**
+   * Sets up a map with passed options and calls a corresponding methods for creating markers
+   *
+   * @method linkRef
+   * @param {Object} ref - node object to be linked as a ref; a map will be rendered here
+   */
+  linkRef = (ref) => {
```
asci: why arrow function?

components/googleMap/googleMap.md
```diff
@@ -0,0 +1,138 @@
...
+      enableTestCustomMarker: false,
+      points: [{
+        position: { lat: 52.379189, lng: 4.899431 },
+        customMarker: true,
+        info: {
+          title: 'custom title',
+        },
+      }],
+      center: { lat: 52.379189, lng: 4.899431 },
+      testCustomStyles: [
```
asci: why it is called `__test__CustomStyles`?

components/googleMap/googleMap.scss
```diff
@@ -0,0 +1,26 @@
+.ui-google-map__custom-marker {
```
asci: could be just `css` file?

components/googleMap/googleMap.js
RumataDeEstor: @asci as a short alternative to `bind`

components/googleMap/googleMap.md
RumataDeEstor: @asci I can't get the problem with that

components/googleMap/googleMap.scss
RumataDeEstor: Why? As I see atm we have `.scss` everywhere

asci: not everywhere https://github.com/Travix-International/travix-ui-kit/blob/7f5d9bfdb31a0cae9550a2d4a370a5f170d42a49/components/index.scss#L6

we're possible will move out of SCSS, so if you don't use any significant features — why not to start use plain CSS just now?

components/googleMap/googleMap.md
asci: just wondering word `test` in props. Was it made for testing only?

RumataDeEstor: @asci I tried to mark that it is only example, not the exact implementation of marker / styles. But np, I removed word `test`, just put '(example)' there.

components/googleMap/googleMap.js
asci: Oh, ok, got it, thanks! Wasn't aware that is it already on stage 2

components/googleMap/googleMap.js
AlleeX: @asci done. Thanks!

components/googleMap/googleMap.js
AlleeX: @asci done. Thanks! ;)

components/googleMap/googleMap.js
AlleeX: @asci done. Thanks!

components/googleMap/googleMap.scss
AlleeX: @asci Because in this file we are use '&'

asci: yes, but you can easily avoid using it and this will make file even better (with pure flat structure)

AlleeX: `.ui-google-map__custom-marker {
...
}

.ui-google-map__custom-marker:before {
...
}

.ui-google-map__custom-marker:after {
...
}`

@asci Do you mean Is it more pure flat structure (copypaste class names) ??

Then why did we start using sass before? :)

AlleeX:
@asci all your remarks has been fixed. Thanks a for for review!)

components/googleMap/googleMap.scss
asci: because we used SASS variables, now we're moving to CSS Variables

AlleeX:
@mAiNiNfEcTiOn regarding sass... but in the PR, which was merged today, guys used sass. Strange... but ok. I'll change

AlleeX:
@asci @mAiNiNfEcTiOn converting .scss to .css has been done

asci:
regarding the PR, which was merged today — I didn't notice, my bad.

mAiNiNfEcTiOn:
@AlleeX indeed, must've missed it. Not personal, though 😉

mAiNiNfEcTiOn:
@asci Changes requested were done, I'm merging.

PR #139, opened 2017-08-09
Adds link variation to button

froskie:
## What this PR do
- Adds link variation to button
image

- Diffs

### DIFF
--- a/components/button/button.scss
+++ b/components/button/button.scss
@@ -101,22 +101,21 @@
 }
 
 .ui-button_variation_link {
-  display: inline-block;
   background: transparent;
   border: none;
-  padding: 0;
-  line-height: normal;
-
   color:  var(--tx-button-link-color);
-  text-decoration:  var(--tx-button-link-text-decoration);
+  display: inline-block;
   font-family: var(--tx-button-link-font-family);
   font-size: var(--tx-button-link-font-size);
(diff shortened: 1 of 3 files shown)


components/button/button.scss
```diff
@@ -99,3 +99,24 @@
   font-size: 28px;
   padding: 15px 20px;
 }
+
+.ui-button_variation_link {
+  display: inline-block;
+  background: transparent;
```
yurist38: I guess, we have an agreement to sort alphabetically css props...

froskie: You're right

themes/_default.yaml
```diff
@@ -145,6 +145,21 @@ button:
     font-family: "var(--tx-generic-font-primary-font-family), var(--tx-generic-font-primary-generic-family)"
     hover-color: *primary
     hover-border-color: *primary
+  link:
+    color: *text
+    font:
+      family: "var(--tx-generic-font-secondary-font-family)"
+      size: "14px"
+      weight: "var(--tx-generic-font-primary-weight-regular)"
```
asci: can't you use these variables directly?

froskie: By directly you mean the actual values? The way it is adds more flexibility per theme.

PR #146, opened 2017-08-16
Change slidingPanel close button to use 'data-rel' attr instead of 'rel'

yurist38:
## WHAT DOES THIS PR DO: ##

* Changed `rel` attribute in close button to `data-rel` to be able to make custom close button with `<Button dataAttrs={({ rel: 'close' })} />` component.

* Diff

## UNIT TESTS: ##

* Adapted to pass 100%

### DIFF
--- a/components/slidingPanel/slidingPanel.js
+++ b/components/slidingPanel/slidingPanel.js
@@ -79,7 +79,7 @@ export default class SlidingPanel extends Component {
 
     this.panel.addEventListener('transitionend', this.handleTransitionEnd);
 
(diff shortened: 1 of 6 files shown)


mAiNiNfEcTiOn:
Before I can approve this, you must explain to me what is this fixing?

yurist38:
@mAiNiNfEcTiOn Problem is not related to styling, it's about functionality. In `SlidingPanel` we can use `<button rel="close">` in any place to close the panel. But it's not achievable right now because there is no way to pass `rel` attribute from `<Button>` component to `<button>` element. But we can easily pass `data-` attributes. Have I answered your question?

PR #148, opened 2017-08-16
Change dependencies type

JutIgor:
In case when it is necessary to use `builder` like an external projects build tool, `extract-text-webpack-plugin`, `sass-loader`, `webpack` need to be installed (because `build` requires them). devDependencies are not installing for packages. That's why these packages were moved to dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,10 @@
     }
   ],
   "dependencies": {
-    "react-select": "^1.0.0-rc.5"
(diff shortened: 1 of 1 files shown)


asci:
@JutIgor Hello, could you please describe what is your case?

fahad19:
I agree with @mAiNiNfEcTiOn. 

I don't think this PR should be merged. this will cause us more trouble in future, especially when your own application's dependencies always need to match with the deps of this package.

PR #155, opened 2017-09-04
Tooltip component

RumataDeEstor:
#### WHAT DOES THIS PR DO:
* added Tooltip component

selection_006

#### WHERE SHOULD THE REVIEWER START:
Diff
#### UNIT AND/OR FUNCTIONAL TESTS:
passed


mAiNiNfEcTiOn:
@RumataDeEstor please add a description.

components/_helpers.js
```diff
@@ -67,10 +67,15 @@ function normalizeDate(dateObject, hours = 0, minutes = 0, seconds = 0, millisec
   return dateObject;
 }
 
+function getNum(str) {
```
maximuk: imho we don't need this function.
And better name is `parseInt10` if you will decide to keep it

components/overlayTrigger/overlayTrigger.js
```diff
@@ -0,0 +1,155 @@
...
+    const actions = {
+      click: this.getOnClickTarget,
+      hover: this.getOnHoverTarget,
+    };
+
+    const targetElement = actions[triggerAction].call(this);
+
+    const elemToToggle = React.cloneElement(this.props.elemToToggle, {
+      active: this.state.active,
+      onCloseButtonClick: this.handleCloseButtonClick.bind(this), // eslint-disable-line
```
maximuk: why are bind it here?

components/overlayTrigger/overlayTrigger.js
```diff
@@ -0,0 +1,155 @@
...
+
+    onElementShow && onElementShow();
+  }
+
+  hideElement() {
+    const { onElementHide } = this.props;
+
+    this.setState({ active: false });
+
+    onElementHide && onElementHide();
```
mAiNiNfEcTiOn: Why not adding a:

```js

static defaultProps = {
  onElementHide = () => {},
  onElementShow = () => {},
}

```

This would allow you to do:

```onElementHide()```

(same for show)

File: components/overlayTrigger/overlayTrigger.js [hunk 4]
```diff
@@ -0,0 +1,155 @@
...
+
+  render() {
+    const { triggerAction } = this.props;
+
+    const actions = {
+      click: this.getOnClickTarget,
+      hover: this.getOnHoverTarget,
+    };
+
+    const targetElement = actions[triggerAction].call(this);
```
mAiNiNfEcTiOn: This is the same as binding.

If you don't want to set the `this.getOnClickTarget = this.getOnClickTarget.bind(this);` in the `constructor`, at least do the `getOnClickTarget = () => { /* ... */ }` to not reset the context on render ;)

components/overlayTrigger/overlayTrigger.js
```diff
@@ -0,0 +1,155 @@
...
+    );
+  }
+}
+
+OverlayTrigger.propTypes = {
+  /**
+   * The content which triggers the interactive component
+   */
+  children: PropTypes.oneOfType([
+    PropTypes.node,
```
mAiNiNfEcTiOn: I think this type already supports the other 2, no?

components/overlayTrigger/overlayTrigger.js
AlleeX: done

components/overlayTrigger/overlayTrigger.js
AlleeX: done

components/overlayTrigger/overlayTrigger.js
AlleeX: Yep. You're correct. Thanks. Fixed

PR #163, opened 2017-09-05
Implement Toggle button component

fope:
Implement ToggleButton component.

- created component;
- created css styling;
- created md file;
- created unit tests

togglebutton1

screen shot 2017-09-05 at 7 53 48 pm


components/toggleButton/toggleButton.scss
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    &_active {
+      background-color: var(--tx-generic-color-active);
+      border-color: var(--tx-generic-color-active);
+      color: var(--tx-generic-color-blank);
+  
+      &::before {
+        color: var(--tx-generic-color-blank);
+        content: '✅';
```
yurist38: Should we use ASCII code instead? It doesn't match design right now.

components/toggleButton/toggleButton.js
```diff
@@ -0,0 +1,94 @@
...
+          onClick={this.handleSelectItem}
+        >
+          {item}
+        </li>
+      );
+    });
+  }
+
+  render() {
+    if (!(Array.isArray(this.props.items) && this.props.items.length > 1)) {
```
yurist38: Would be nice to destructure `items` from `this.props` before this line. It will give a bit more readability. And also using negation separately would simplify a bit.
```
const { items } = this.props;
if (!Array.isArray(items) || !items.length)) {
   ...
}
```
this way looks more readable for me.

yurist38: We even could remove check for array if use default value.

components/toggleButton/toggleButton.scss
fope: Done.

components/toggleButton/toggleButton.js
fope: Done.

components/toggleButton/toggleButton.js
```diff
@@ -0,0 +1,97 @@
...
+   * @method handleSelectItem
+   * @param {Object} event
+   * @private
+   */
+  handleSelectItem = (event) => {
+    event.stopPropagation();
+
+    const index = this.props.items.indexOf(event.target.textContent);
+    this.setState({ activeItem: index });
+    (typeof this.props.handleSelect === 'function') && this.props.handleSelect(index);
```
mAiNiNfEcTiOn: I'd only check if is set:

```js
this.props.handleSelect(index);
```

Since, based on the PropTypes it shouldn't be anything but a function and it's even required. 

If it wasn't required, you could also define it, as a defaultProp being a no-op, which would then make this easier:

```js
class ToggleButton extends React.Component {
  static defaultProps = {
    handleSelect = () => {},
  }
  
  /* ... */
  handleSelectItem = (event) => {
    /* ... */
   this.props.handleSelect(index);
   }
}
```

Still, looks good as it is already :) 👍  nice work

components/toggleButton/toggleButton.js
```diff
@@ -0,0 +1,97 @@
...
+import { getClassNamesWithMods } from '../_helpers';
+
+/**
+ * ToggleButton component.
+ */
+class ToggleButton extends Component {
+  /**
+   * Initialize default state
+   */
+  state = {
```
asci: I would prefer to make it stateless and control selected item from outside. With current implementation, it is not possible to pre select item by default

PR #164, opened 2017-09-05
Move react-select from devDeps to deps

yurist38:
## WHAT DOES THIS PR DO: ##

* `react-select` package moved back to dependencies. Since we moved it out of it few errors have appeared (on tests and pipelines).

* Diff

## UNIT TESTS: ##

* Passed

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,30 +58,29 @@
     "coveralls": "^2.11.12",
     "create-react-class": "^15.5.2",
     "css-loader": "~0.28.1",
-    "enzyme-to-json": "^1.1.4",
     "enzyme": "^2.7.1",
+    "enzyme-to-json": "^1.1.4",
+    "eslint": "^3.14.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -95,5 +94,8 @@
       "sass-loader",
       "webpack"
     ]
+  },
+  "dependencies": {
```
fahad19: common practice to keep `dependencies` above `devDependencies`.

yurist38: That's strange because it was done automatically by `npm`

yurist38: Actually, I don't know how exactly `npm` sorting deps and devDeps. Maybe it depends on which one was added first. But it will make more sense if `npm` has a strict order for them. In this case better to follow that order.

yurist38: Tested it now, seems like `npm` just adding new section into the end of file. So I will move deps to position before devDeps.

PR #171, opened 2017-09-11
Adds 'set -e' to .travis.yml

mAiNiNfEcTiOn:
# What does this PR do:

  - Adds the flag `set -e` to the build process so that it stops the build as soon as it finds an error.

  - Check the failing build being stopped at `ci:lint` task: https://travis-ci.org/Travix-International/travix-ui-kit/jobs/274109638

Since there's another issue with the CI process, it is expected that the build fails 😉

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@ language: node_js
 node_js:
   - '6'
 
+before_install:
+  - set -e
+
 script:
   - 'npm run ci:lint'
   - 'npm run ci:test'


.travis.yml
```diff
@@ -3,6 +3,9 @@ language: node_js
 node_js:
   - '6'
 
+before_install:
+  - set -e
```
jackTheRipper: wouldn't it be cool to have the `-x` enabled if the env var `DEBUG` is equal to one? Or aren't we expecting much bash logic in the the CI?

mAiNiNfEcTiOn: @jackTheRipper nope, we're not. This is purely execution of commands without logic.

But can you explain a little bit more in detail what `-x` would allow us to do?

mAiNiNfEcTiOn: Oh well, Google is our friend, here's some more info: https://stackoverflow.com/questions/36273665/what-does-set-x-do

mAiNiNfEcTiOn: @jackTheRipper however, the commands executed at CI level are already printed on Travis CI.

Nevertheless, I'm all open for adding it at a later stage if you see a benefit on it!

jackTheRipper: @mAiNiNfEcTiOn sorry for the late reply. I offered to add the `-x` flag in debug mode for having the details of the bash commands. But such as what I said in my original comment it makes sense only if we do have some logic in bash (which seems to be not the case for this project currently)

PR #189, opened 2017-09-19
Sliding panel improvements

RumataDeEstor:
# What does this PR do:

* adds props to be able to call passed function before closing starts and prevent closing if it returns false;
* adds props for customizing left / right blocks in the header (custom buttons, etc.);

   components/slidingPanel/slidingPanel.js

### DIFF
--- a/components/slidingPanel/slidingPanel.js
+++ b/components/slidingPanel/slidingPanel.js
@@ -11,7 +11,7 @@ export default class SlidingPanel extends Component {
 
     this.handleClickOverlay = this.handleClickOverlay.bind(this);
     this.handleActive = this.handleActive.bind(this);
-    this.handleClickCloseButton = this.handleClickCloseButton.bind(this);
+    this.handleClose = this.handleClose.bind(this);
     this.handleTransitionEnd = this.handleTransitionEnd.bind(this);
   }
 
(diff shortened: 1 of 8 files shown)


components/slidingPanel/slidingPanelHeader.js
```diff
@@ -1,28 +1,51 @@
...
+  const defaultCloseButton = (
+    <button
+      className="ui-sliding-panel-header__close-button"
+      data-rel="close"
+    >
+      &#215;
+    </button>
+  );
+
+  const left = leftBlock
```
froskie:
```
const left = leftBlock || (
            <div className="ui-sliding-panel-header__left-block">
              {leftBlock}
            </div>
          );
```

components/slidingPanel/slidingPanel.js
```diff
@@ -27,6 +27,12 @@ export default class SlidingPanel extends Component {
     closeOnOverlayClick: PropTypes.bool,
 
     /**
+     * Hook that will be executed when trying to close a panel if exists.
+     * If it returns false, the panel won't be closed.
+     */
+    beforeClosing: PropTypes.func,
```
froskie: Although the whole code is ok, I'd say that it'd sound better if this method was named `shouldClose` instead of `beforeClosing`, because that name implies a method called before close, nothing else.

RumataDeEstor: @froskie had a huge discussion here in Brest, we agreed it that 1) it's like typical hooks: before / after 2) it not just defines should component be closed or not, but also implement custom logic when trying to close. 

I also was thinking about *onTryingToClose*

PR #203, opened 2017-10-13
Remove wrong attribute from svg

DzmitryKukharuk:
# What does this PR do:
   Spinner component has a wrong attribute "styles". This attr has been removed.

   Please see the diff


asci:
why remove? Why not fix? also, could you please add a picture of the spinner (using non white background)?

DzmitryKukharuk:
image

The spinner with my changes looks totally the same as without. It means that "styles" attr is redundant and does not impact spinner`s style.

AlleeX:
@DzmitryKukharuk did you check not others background colors? Can you please add example with (for example)  blue background to the docs

Like this

selection_122


Thanks ;)

DzmitryKukharuk:
Hi Alleex,
sure, I checked with diff colors.
This is an example with blue color:
image

PR #204, opened 2017-10-19
Button brand fixes

MadinaShad:
Fixed buttons on travix UI-kit
screen shot 2017-10-23 at 11 00 18

### DIFF
--- a/components/button/button.md
+++ b/components/button/button.md
@@ -6,7 +6,7 @@ Basic button:
       <Button dataAttrs={{gtm: 'some-id'}} type="reset">With GTM id</Button><br/><br/>
       <Button mods={['test']} type="reset">With mode</Button><br/><br/>
       <Button mods={['test']} className="my-class">Example with custom class set</Button><br/><br/>
-      <Button onMouseUp={alert(onMouseUp)}>Example with custom event (onMouseUp)</Button><br/><br/>
+      <Button onMouseUp={() => alert('Mouse Up')}>Example with custom event (onMouseUp)</Button><br/><br/>
     </div>
 
 Sizes:
--- a/components/button/button.md
+++ b/components/button/button.md
@@ -25,6 +25,9 @@ Variations - Ghost:
       <Button onClick={alert} variation="ghost" size="m">Medium Ghost</Button><br/><br/>
       <Button onClick={alert} variation="ghost" size="l">Large Ghost</Button><br/><br/>
       <Button onClick={alert} variation="ghost" size="xl">Extra large Ghost</Button><br/><br/>
+      <div style={{background: "#005FBB", padding: "10px", display: "inline-block"}}>
+        <Button onClick={alert} variation="ghost-inverted" size="s">Small Ghost Inverted</Button><br/><br/>
+      </div>
     </div>
 
 Variations - Link:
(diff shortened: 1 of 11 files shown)


themes/_default.yaml
```diff
@@ -114,37 +114,45 @@ badge:
...
-      active: "var(--tx-generic-color-primary-lighter)"
-  color:
-    bg-top: "var(--tx-generic-color-accent)"
-    bg-bottom: "var(--tx-generic-color-accent)"
-    text: "var(--tx-generic-color-blank)"
-    border: "var(--tx-generic-color-accent-dark)"
-    shadow: "transparent"
+  default:
+    normal:
+      background: "var(--tx-button-default-normal-background)"
```
iwwwi: please adjust the changes as we discussed ;) 
also check other similar definitions :)

components/button/button.md
```diff
@@ -27,6 +27,9 @@ Variations - Ghost:
       <Button onClick={alert} variation="ghost" size="m">Medium Ghost</Button><br/><br/>
       <Button onClick={alert} variation="ghost" size="l">Large Ghost</Button><br/><br/>
       <Button onClick={alert} variation="ghost" size="xl">Extra large Ghost</Button><br/><br/>
+      <div style={{background: "#005FBB", padding: "10px", display: "inline-block"}}>
```
iwwwi: what is the purpose to hard code the styles here?

AlleeX: @iwwwi this is just for documentation. 

selection_123

themes/_default.yaml
MadinaShad: yep! Thanks, Ivana))

components/button/button.md
MadinaShad: maybe because those styles are not related to the button components. It's background under a button which shows how button will look like on a dark background @froskie what do you think? Should I move it to sass file?

iwwwi: @AlleeX @MadinaShad cool :) doesn't have to remove it ;)

themes/_default.yaml
```diff
@@ -22,37 +24,38 @@ generic:
...
+    secondary-lighter: &secondary-lighter       "#FFD97C"
 
     text-dark: &text-dark "#333"
     text: &text "#666"
     text-light: &text-light "#999"
     text-lighter: &text-lighter "#998"
     text-lighter-rgb: &text-lighter-rgb "153,153,136"
 
-    active: &active "#3892e7"
+    active: &active                             "#4baf93"
```
froskie: I think something went wrong here   =]

themes/_default.yaml
```diff
@@ -1,5 +1,7 @@
 generic:
   border:
+    radius-default-btn: &radius-default-btn 2px
+    radius-ghost-btn: &radius-ghost-btn 2px
```
froskie: I was looking for the usage of those. Could you point me out where you're using this?

MadinaShad: download

themes/_default.yaml
MadinaShad: @froskie depending on what alignment standards you use, some variables align by colors, some variables no) for example: 
screen shot 2017-10-23 at 15 57 42
How it should be aligned?

themes/_default.yaml
```diff
@@ -113,6 +114,44 @@ badge:
   padding:
     right-left: 8px
     top-bottom: 6px
+ui-button:
```
asci: I'm sorry, I missed this — what was the reason to add `ui-button` root, instead of fixing `button`?

AlleeX: @asci I also have the same question. This is answer from @froskie: "Because of backward compatibility. Everything related to the actual button component on UI Kit should be `.ui-button`"

asci: but it is not backward compatible because those variables are used only in `button.scss` file, which is basically the only file for styles for button

PR #205, opened 2017-10-19
Adds the 'intialSelectedIndex' attribute to ToggleButton

mAiNiNfEcTiOn:
# What does this PR do:
  * Adds the `intialSelectedIndex` prop which allows the user to define the _item selected by default when mounting_.

  * Diffs
  * Styleguide

### DIFF
--- a/components/toggleButton/toggleButton.js
+++ b/components/toggleButton/toggleButton.js
@@ -7,11 +7,15 @@ import { getClassNamesWithMods } from '../_helpers';
  * ToggleButton component.
  */
 class ToggleButton extends Component {
-  /**
-   * Initialize default state
-   */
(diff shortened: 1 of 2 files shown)


mAiNiNfEcTiOn:
Immediately triggering discussion with the y'all...

This component doesn't have the `componentWillReceiveProperties`, which means if you change the props on the documentation, it doesn't react to it 😉  😄 

Should I change it to instead have a `selectedItemIndex` property, change it to be _stateless_ and depend only on props?

#yourThoughts

asci:
I prefer for stateless components

mAiNiNfEcTiOn:
Ok, will change it then ... :) Thanks for your feedback

PR #206, opened 2017-10-19
Change ToggleButton component to a functional, stateless component

mAiNiNfEcTiOn:
# What does this PR do:
  * Converts the `ToggleButton` into a stateless component.
  * Updates the documentation/styleguide to have examples.
  * Implements the tests correctly.

  - Diffs
  - Check the styleguide.
  - Take a look at how the tests are done. It should be as simple as is.


asci:
Looks like it is not back compatible. Will changes to projects using this follow? Or we'll change major version?

mAiNiNfEcTiOn:
@asci unless we've already reached 1.0, I was unset tyher impression that minors were our "current" majors... And this is tagged as minor

components/toggleButton/toggleButton.js
```diff
@@ -1,85 +1,49 @@
...
+    handleSelect,
+    items,
+    selectedIndex,
+  } = props;
 
-    const index = this.props.items.indexOf(event.target.textContent);
-    this.setState({ activeItem: index });
-    (typeof this.props.handleSelect === 'function') && this.props.handleSelect(index);
+  if (!Array.isArray(items) || items.length < 2) {
+    return <noscript />;
```
spzm: What react version should we support? If it's 15 then we can start using `null`.

mAiNiNfEcTiOn: Good point. However, the purpose of this PR it's to convert the component to stateless. That was based on the master which still uses R14 on the package.json

Nevertheless, if someone is taking care of that,  I can wait, rebase and change the code.

mAiNiNfEcTiOn: Ok I took care of it 🙄  and will change that.

mAiNiNfEcTiOn:
@Travix-International/frontend @Travix-International/frontend-epam changed to patch since no one is using it on any project 🙄

PR #209, opened 2017-10-24
Add icon prop to dropdown

JutIgor:
# What does this PR 

* Changes the `DropDown` component to support the `icon` attribute that affects the display of value label. 

dd1
dd2

* Diffs

### DIFF
--- a/components/dropDown/dropDown.js
+++ b/components/dropDown/dropDown.js
@@ -101,6 +101,7 @@ class DropDown extends Component {
       clearable,
       dataAttrs = {},
       filterMode,
+      icon,
       mods = [],
       multi,
       name,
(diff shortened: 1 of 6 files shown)


components/dropDown/dropDown.js
```diff
@@ -97,6 +110,7 @@ class DropDown extends Component {
       searchable,
       status,
       value,
+      icon,
```
yurist38: Not sorted alphabetically.

components/dropDown/dropDown.js
```diff
@@ -198,6 +213,10 @@ DropDown.propTypes = {
  * Initial field value
  */
   value: PropTypes.any,
+  /**
```
yurist38: Please sort it here as well

yurist38:
@JutIgor Could you please add some screenshot of that what have been added? Thanks!

PR #212, opened 2017-10-27
Tabs component

AlleeX:
# What does this PR do:

    Tabs component

vokoscreen-2017-10-27_11-14-30

    Diff


components/tabs/tab.js
```diff
@@ -0,0 +1,129 @@
...
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+
+import { getClassNamesWithMods, getDataAttributes } from '../_helpers';
+
+/**
+ * Tab component
+ */
+class Tab extends Component {
+  handleTabClick = (event) => {
```
froskie: Could you let me know in which proposal stage this is?

AlleeX: @froskie ?

AlleeX: selection_164

components/tabs/tab.js
```diff
@@ -0,0 +1,129 @@
...
+    const { value } = this.props;
```
yurist38: How about destructuring `onClick` as well as `value`?

components/tabs/tab.js
spzm: stage 3 as I see https://github.com/tc39/proposal-class-fields

tests/unit/tabs/tabs.spec.js
```diff
@@ -0,0 +1,118 @@
+import { shallow, mount } from 'enzyme';
+import React from 'react';
+import Tabs from '../../../components/tabs/tabs';
+import Tab from '../../../components/tabs/tab';
+
+describe('Radio Button', () => {
```
mAiNiNfEcTiOn: :heart:

PR #213, opened 2017-10-29
Add prop renderEmptyItems to the List component

yurist38:
# What does this PR do:

    Added prop `renderEmptyItems` to the `List` component. Value by default is `true`.

    Diff

### DIFF
--- a/components/list/list.js
+++ b/components/list/list.js
@@ -7,7 +7,13 @@ import { getClassNamesWithMods } from '../_helpers';
  * General List component. Use when you need to display array of elements
  */
 function List(props) {
-  const { items, hideBullets, align } = props;
+  const {
+    align,
+    hideBullets,
+    items,
+    renderEmptyItems,
+  } = props;
+
   const mods = props.mods ? props.mods.slice() : [];
 
   mods.push(`align_${align}`);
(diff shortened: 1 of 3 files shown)


components/list/list.js
```diff
@@ -32,33 +42,39 @@ function List(props) {
 }
 
 List.defaultProps = {
-  hideBullets: false,
   align: 'vertical',
+  hideBullets: false,
+  renderEmptyItems: true,
```
asci: any case when it should be `true` by default?

yurist38: Not sure that there is any. I did that way just to save existing functionality.

components/list/list.js
```diff
@@ -18,7 +24,11 @@ function List(props) {
 
   const className = getClassNamesWithMods('ui-list', mods);
 
-  const itemsBlock = items.map((item, index) => (
+  const notEmpty = i => i !== '' && i !== null;
+
+  const itemsToRender = (renderEmptyItems === false) ? items.filter(notEmpty) : items;
```
mAiNiNfEcTiOn: Good one. You could've also filter by `Boolean`:

```js
items.filter(Boolean)
```

yurist38: But it that case we will get 0 values filtered out, I wanted to prevent that. @mAiNiNfEcTiOn

PR #217, opened 2017-10-31
Updates React (and it's siblings) to v15

mAiNiNfEcTiOn:
# What does this PR do:
  * Updates `react` and it's related packages to v15.

  * Diffs


mAiNiNfEcTiOn:
@Travix-International/frontend  @Travix-International/frontend-epam I've removes all labels since `react`and it's fellow packages are not `dependencies` (but `devDependencies`) and this change does not affect the resulting package (yet).

PR #222, opened 2017-11-07
Upgraded enzyme to version 3, fixed tests

### LINKED ISSUE #218: Upgrade `enzyme` from 2.x to 3.x

Not applicable

# Description

When updating `enzyme` to version 3.x it breaks a lot of tests, requires a new dependency -
 `enzyme-adapter-react-15` - and a setup file (to be configured in the `tests/jest.config.json`).

Besides that it requires changes on the tests. In cases where the state is being changed via lifecycle methods, rather than simulated events, you now have to explicitly do:

```js
 // Imagine that on MyComponent, the componentDidMount lifecycle runs this.setState()
const wrapper = mount(<MyComponent);
wrapper.update();
```
Otherwise `enzyme` doesn't know if the state changed (this is one of the breaking changes between 2.x and 3.x.

# Steps to reproduce

Update `enzyme` to the latest version and try to run the tests. It immediately tells you that you need the adapter 😄 

# Expected results

Tests passing 😉 

# Actual results

Tests not passing.

> cc'ing @Travix-International/frontend  and @Travix-International/frontend-epam

yurist38:
# What does this PR do:
- Migrate from `enzyme` v.2 to v.3. Closes #218 

- Diff

# What was changed according to the new `Enzyme` version:
- It requires adapter now. That means one more devDependency `enzyme-adapter-react-15` and some extra configuration. A common practice to configure it in the separate file, I followed that way (`tests/setupTests.js`)
- Seems like behavior of `simulate` method was changed. Now when we simulate a click it doesn't rerender the tree automatically and we need to render it manually to check the changes after click. In some cases you will need to call `.update()` method from root wrapper or call `.render()` method from the node which should be updated.
- If component should render `null` we can test it by checking his `.type()` and compare it to `null`. No need to match snapshot.
- When you need to access direct children nodes this way `.find('> span')` doesn't work anymore. You can use `.children('span')` instead.
- Now use `.instance()` instead of `.getNode()`

[Migration guide](http://airbnb.io/enzyme/docs/guides/migration-from-2-to-3.html)


mAiNiNfEcTiOn:
@yurist38 I know we can read the code, but would it be possible to add to the PR what are the main differences in testing that we should be aware of, due to the switch to the new major enzyme version?

yurist38:
@mAiNiNfEcTiOn that's right, my bad. I will add more information to PR's description.

PR #223, opened 2017-11-08
Sliding panel improvements

RumataDeEstor:
# What does this PR do:

Provides the following improvements for Sliding Panel:

- Adds prop to define the direction of the panel.
- Adds optional *subheader* block without paddings to display the additional info in the top under the header.
- Adds optional prop to display the default block with back button with arrow in the left side of the panel .
- Adds prop to change the width of the panel.
- *Title* now can be a node.
- No more default bold font for the whole panel body.
- "X" button is smaller according common designs. 
- Fix header styles to make the title always be centered relative to the width of the header block.

components/slidingPanel/slidingPanel.js


components/slidingPanel/slidingPanelHeader.scss
```diff
@@ -22,17 +24,55 @@
   color: var(--tx-generic-color-secondary-darker);
   cursor: pointer;
   font-family: Arial, sans-serif;
-  font-size: 40px;
-  font-weight: 400;
+  font-size: 32px;
```
RumataDeEstor: vars

components/slidingPanel/slidingPanel.js
```diff
@@ -27,6 +27,22 @@ export default class SlidingPanel extends Component {
     closeOnOverlayClick: PropTypes.bool,
 
     /**
+     * When true, it will show the block with arrow icon and passed text (optional).
+     * You can either enable it, or use leftBlock property to have more customization.
+     */
+    useDefaultLeftBlock: PropTypes.bool,
```
RumataDeEstor: sort

components/slidingPanel/slidingPanel.js
```diff
@@ -164,13 +199,35 @@ export default class SlidingPanel extends Component {
 
   render() {
     const {
-      dataAttrs,
       children,
-      title,
+      dataAttrs,
+      direction,
       leftBlock,
```
RumataDeEstor: sort

components/slidingPanel/slidingPanel.js
```diff
@@ -164,13 +199,35 @@ export default class SlidingPanel extends Component {
...
+      width,
+      useDefaultLeftBlock,
+      backButtonLabel,
+      onBackButtonClick,
     } = this.props;
 
+    const headerLeftBlock = useDefaultLeftBlock
+      ? (
+        <span>
+          <button
```
RumataDeEstor: extract

components/slidingPanel/slidingPanel.js
```diff
@@ -68,13 +89,27 @@ export default class SlidingPanel extends Component {
     active: PropTypes.bool,
 
     /**
+     * Defines the width of the panel.
+     */
+    width: PropTypes.string,
```
asci: why do we need this prop?

RumataDeEstor: we have different width for panels (I found at least 2 different types). If we will have this prop, we will not need to override the width in styles

components/slidingPanel/slidingPanel.js
```diff
@@ -27,54 +37,80 @@ export default class SlidingPanel extends Component {
...
+     * When true, it will show the block with arrow icon and passed text (optional).
+     * You can either enable it, or use leftBlock property to have more customization.
+     */
+    useDefaultLeftBlock: PropTypes.bool,
+
+    /**
+     * Defines the width of the panel.
      */
-    title: PropTypes.string,
+    width: PropTypes.string,
```
asci: - is it possible to have here standard sizes?
- how will it work on mobile?

yurist38:
@asci let's discuss it once you're here. About mobiles - it works fine there, when width prop is wider than screen it will show on full screen width, same behavior as we had on mobiles before.

Issue #194, opened 2017-10-03
Extend Price component to support ability to add symbols after the price


IvanPresmytsky:
**Browsers and versions affected**

n/a

**Description**

In some cases there is need to add asterisk or some other symbol after Price component.

**Steps to reproduce**

![asterisk-needed-to-ui-kit-price](https://user-images.githubusercontent.com/12682710/31127743-5cb2bc04-a858-11e7-937e-6028fd516873.png)

**Expected results**

Have ability to add asterisk or other symbol to the Price component

Issue #187, opened 2017-09-15
Extend Price component to support vertical align customization for its items


IvanPresmytsky:
# Browsers and versions affected

n/a

# Description
In some cases there is need to customize vertical align of Price component item. For example price currency. It depends on Prices font type. 

# Steps to reproduce
it's default behavior for Price component with "Khand" font family. (default "align-items" - "flex-start")
![otherdays-widget-price-component-align-items-flex-start](https://user-images.githubusercontent.com/12682710/30478444-b7e1ea5c-9a19-11e7-9d34-d3cac6cd9b7c.png)

# Expected results

Have ability to customize vertical align of Price component items.

# Actual results
![otherdays-widget-price-component-align-items-flex-start](https://user-images.githubusercontent.com/12682710/30478372-74dc30d2-9a19-11e7-9c9f-6be48b75e8c1.png)

PR #224, opened 2017-11-09
Update price with asterisk and vertical align

IvanPresmytsky:
# What does this PR do:

   * Extend Price component to support vertical align customization for its items by line-height variables
      Fixed issue #187 

   * Extend Price component to support ability to add symbols after the price (asterisk)
      Fixed issue #194 

   * Modify horizontal margins between currency, integers and decimals.

   * Modify underline height for small and extra small states

    Diffs

Cheaptickets

extendend-price-component cheaptickets

Budgetair

extending-price-component budgetair -after


components/price/price.md
```diff
@@ -22,9 +29,9 @@ Price with discount:
 Sizes:
 
     <div>
-      <Price size="xs" value={50153.30} /><br/><br/>
-      <Price size="s" value={50153.30} /><br/><br/>
-      <Price size="m" underlined value={50153.30} /><br/><br/>
-      <Price size="l" underlined value={50153.30} /><br/><br/>
-      <Price size="xl" underlined value={50153.30} /><br/><br/>
+      <Price showAsterisk size="xs" underlined value={50153.30} /><br/><br/>
+      <Price showAsterisk size="s" underlined value={50153.30} /><br/><br/>
```
iwwwi: I would suggest to leave few examples without asterisk or add both cases of sizes (with and without asterisk)

yurist38:
@IvanPresmytsky Thank you for reacting on my message and updating description with screenshot! Maybe I don't understand something but I don't see any asterisk on the it... Is that an old version on the screenshot?

IvanPresmytsky:
@yurist38 
Thanks for remarks.
Updated

yurist38:
Thanks @IvanPresmytsky ! But it looks like space is missing between `€` sign and price number, isn't it?

IvanPresmytsky:
@yurist38 
Updated :)

components/price/price.js
```diff
@@ -100,6 +101,15 @@ function Price(props) {
       )
     : null;
 
+  const asteriskSymbol = '*';
```
froskie: What's the purpose of creating this?

IvanPresmytsky: @froskie 
May be in the future we will need another symbol/symbols. :-)
But for now this variable is redundant. I agree.
Should I remove it?

PR #226, opened 2017-11-10
Added global mode for sliding panel and some improvements for global component

AlleeX:
# What does this PR do:

This is my suggestion. Why?
Because I want to make the sliding panel component more independent.
As I see during the using sliding panel we always should use global wrapper and make logic for manage not scrollable state for body using mount/unmount hack.

Added global mode for sliding panel and some improvements for global component.

    Diff


components/global/global.js
```diff
@@ -5,23 +5,30 @@ import {
...
   componentDidMount() {
-    if (this.props.noscroll) {
-      global.window.document.body.classList.add('ui-global_noscroll');
-    }
-
     this.target = global.window.document.createElement('div');
     this.target.classList.add('ui-global');
     global.window.document.body.appendChild(this.target);
-    this.componentDidUpdate();
+    this.componentDidUpdate({ noscroll: false });
```
mAiNiNfEcTiOn: I know that it was already like this, but can you extract `componentDidUpdate`'s logic into another function and then both `componentDidMount` and `componentDidUpdate` make use of the same code?

Otherwise looks like it's using React's lifecycle methods for custom purpose (which currently it is...)

mAiNiNfEcTiOn:
@AlleeX the changes you made on the SlidingPanel are really necessary?

AFAIK you have the `Global` component exported and you can use it to wrap the component no?

https://github.com/Travix-International/travix-ui-kit/blob/master/components/index.js#L12

components/global/global.js
AlleeX: @mAiNiNfEcTiOn sure, for me this is also looks a bit weird. I'll change that ;)

AlleeX:
@mAiNiNfEcTiOn see my comment in the description =)

spzm:
@mAiNiNfEcTiOn As I see it's just a customization. Because it used quite often in production it makes sense to have it out of the box. In other way almost in every place we need to cover it in global...

mAiNiNfEcTiOn:
@spzm while true that it is a customization, it is in fact bringing an implicit usage of it on the code. Just like the `Modal` component which uses it internally. Don't know why, since the Modal's functionality does not depend on it and, as an HOC, the `Global` component brings the optional layered levels.

It's as simple as using `<Global>` in it.

Also I don't agree that the `noscroll` should be something implicit in the Sidebar panel. Why? Because in cases where I have the xxl (or xl?) size on the SRP, I will have a 2 sidebars appearing at the same time the flight results should be scrollable.

spzm:
@mAiNiNfEcTiOn IMHO I don't think that Global is optional behavior... Why? Because it's part of panel functionality. For example, if you don't use global - panel can be overlapped by stacked context. I can suggest use Global by default without option... but because we need to be backward compatible additional prop looks fine for me.

About `noscroll` I think this decision should come from ux designers. A lot of ui issues can happen just because content scrollable. Can we ask what designers for the input?
As  I see for now - we have blur overlay by default, so it means that content shouldn't bet scrollable by default as well. When somebody will need this component for flight results, then component should be customized: add property that remove overlay and add scroll.

mAiNiNfEcTiOn:
Fine by me. Anyway, as I said, that's not a reason to block 😄. Since the other changes are done, I'm more than happy to merge.

@AlleeX thanks for the changes :)

components/slidingPanel/slidingPanel.js
```diff
@@ -280,5 +288,13 @@ export default class SlidingPanel extends Component {
         </div>
       </div>
     );
+
+    return global
```
mAiNiNfEcTiOn: If it is mandatory, then please remove the global attribute and make it used all the time.

components/global/global.js
```diff
@@ -5,38 +5,49 @@ import {
...
-      </div>
-    ), this.target);
+  componentDidUpdate(prev) {
+    this.updateContainerAndBody(prev.noscroll);
   }
 
   componentWillUnmount() {
     if (this.props.noscroll) {
-      global.window.document.body.classList.remove('ui-global_noscroll');
+      setGlobalNoscroll(false);
```
mAiNiNfEcTiOn: Say that you have 2 components on the page:

`Modal` and `SlidingPanel`, both using `Global`, both with `noscroll=true`.

If `Modal` _unmounts_ and `SlidingPanel` is still _mounted_, can you tell me which classes will the `body` DOM Node have?

components/slidingPanel/slidingPanel.js
mAiNiNfEcTiOn: Again, I'd prefer to not maintain the _backwards_ compatibility and just use _global_ for all cases if the assumption is that all should use Global.

components/global/global.js
```diff
@@ -1,44 +1,63 @@
...
-        {this.props.children}
-      </div>
-    ), this.target);
+  toggleGlobalNoscroll(flag) {
+    const body = global.window.document.body;
+    if (!this.isSettedNoScroll) {
+      return;
+    }
+    flag
+      ? body.classList.add('ui-global_noscroll')
```
asci: is it possible to have 2 `Globals` at the same time? like opened modal from the sliding panel? How will it work then?

components/slidingPanel/slidingPanel.js
```diff
@@ -283,5 +284,11 @@ export default class SlidingPanel extends Component {
         </div>
       </div>
     );
+
+    return (
```
asci: this doesn't seem optional nor backward compatible as it is described

components/global/global.js
maximuk: I added an example https://github.com/Travix-International/travix-ui-kit/pull/226/commits/c68674c0048d67a6ac59720ea951a1eb7d8ea480

components/slidingPanel/slidingPanel.js
vasilibrutsky: Ricardo proposed using Global always with SlidingPanel

PR #227, opened 2017-11-13
Fix Unknown Prop Warning

AlleeX:
# What does this PR do:

    Fix Unknown Prop Warning

selection_010

    Diff


mAiNiNfEcTiOn:
@AlleeX could you add a screenshot of the issue please? Thanks 👍

components/autoComplete/autoCompleteItem.js
```diff
@@ -1,5 +1,6 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
+import omit from 'lodash.omit';
```
asci: according to this comment https://github.com/Travix-International/travix-ui-kit/pull/221#discussion_r149354939 better to use `lodash/omit`

components/autoComplete/autoCompleteItem.js
```diff
@@ -29,13 +30,19 @@ class AutoCompleteItem extends Component {
...
-      ...otherProps
     } = this.props;
     const mods = this.props.mods ? this.props.mods.slice() : [];
 
+    const otherProps = omit(this.props, [
+      'code',
+      'index',
+      'isActive',
+      'isTitle',
+      'onClick',
```
asci: why `onClick` removed?

AlleeX:
@asci @mAiNiNfEcTiOn done

components/autoComplete/autoCompleteItem.js
```diff
@@ -29,13 +30,19 @@ class AutoCompleteItem extends Component {
...
       children,
       dataAttrs = {},
-      onClick,
       isActive,
       isTitle,
-      ...otherProps
     } = this.props;
     const mods = this.props.mods ? this.props.mods.slice() : [];
 
+    const otherProps = omit(this.props, [
```
asci: maybe it will not have a huge impact, but what if move this array to top scope and don't create it for each component render?

mAiNiNfEcTiOn: When an array is used on a specific function and that scope ends, it gets cleaned by the GC.

What happens when it's global? 😉  But nevertheless, as you said, shouldn't be a problem

yurist38:
@AlleeX could you please help me to understand the problem? Why were those extra props passed into the components/elements? Should we maybe try to fix this problem in outside UI-KIT scope?

AlleeX:
@yurist38 https://reactjs.org/warnings/unknown-prop.html

yurist38:
@AlleeX I know that and that's not an explaination actually 😄  My question was not about how the unknown props appears but why do we pass those props, what are shouldn't be there? We try to fix just a problem instead of fixing its root. It looks a bit hackish to me. But let me think a bit and check on my local...

yurist38:
I've just opened [UI-KIT Demo](https://travix-international.github.io/travix-ui-kit/#!/AutoComplete) and don't see any warnings there. First thought - problem is not here 😄

asci:
Very valid point, Yuri. The issue comes from our main repo, not from UI Kit.
I also would like to see whitelisting, instead of blacklisting properties

mAiNiNfEcTiOn:
@yurist38 and @asci have you updated to React 16? Because that seems to be a problem coming from R16: https://reactjs.org/warnings/unknown-prop.html

asci:
yep, Ricardo, you're right, there was some misunderstanding from our side. We thought here Alex filter all passed props. But he does filter only valid props.

yurist38:
@mAiNiNfEcTiOn didn't update yet. But problem itself is pretty clear to me now and it's not coming from React 16. It happens because ui-kit components currently allow any props to be passed into them. That approach looks quite tricky to me. I think we should somehow limitate an input.

mAiNiNfEcTiOn:
@AlleeX so, will be doing the _temporary_ helper function that returns an object with only the props (and it's values) that do not match the ones in the `propTypes`, which you then do a spread operator on it, or should I do it?

AlleeX:
@mAiNiNfEcTiOn yep. I'll do that. Thanks a lot for discussion and suggestion! ;)

AlleeX:
@mAiNiNfEcTiOn I updated my implementation

components/_helpers.js
```diff
@@ -67,8 +67,25 @@ function normalizeDate(dateObject, hours = 0, minutes = 0, seconds = 0, millisec
...
+ * @method ejectOtherProps
+ * @param {Object} props conponent props.
+ * @param {Object} propTypes conponent props types.
+ * @return {Object} custom props.
+ */
+function ejectOtherProps(props, propTypes) {
+  return Object.keys(props)
+    .filter(x => Object.keys(propTypes).indexOf(x) < 0)
+    .reduce((prev, item) => {
+      return { ...prev, [item]: props[item] };
```
mAiNiNfEcTiOn: uuuuuu, fancy! 😉

components/_helpers.js
```diff
@@ -67,8 +67,25 @@ function normalizeDate(dateObject, hours = 0, minutes = 0, seconds = 0, millisec
...
+ * Eject real custom users props from props
+ *
+ * @method ejectOtherProps
+ * @param {Object} props conponent props.
+ * @param {Object} propTypes conponent props types.
+ * @return {Object} custom props.
+ */
+function ejectOtherProps(props, propTypes) {
+  return Object.keys(props)
+    .filter(x => Object.keys(propTypes).indexOf(x) < 0)
```
mAiNiNfEcTiOn: Since you went fancy in the line below, you can go fancy here and do:

```js
const propTypesKeys = Object.keys(propTypes);
Object.keys(props)
  .filter(x => propTypesKeys.includes(x))
  // ...
```

AlleeX: @mAiNiNfEcTiOn ``` .filter(x => !propTypesKeys.includes(x)) ``` ;)

components/_helpers.js
```diff
@@ -76,8 +76,9 @@ function normalizeDate(dateObject, hours = 0, minutes = 0, seconds = 0, millisec
  * @return {Object} custom props.
  */
 function ejectOtherProps(props, propTypes) {
+  const propTypesKeys = Object.keys(propTypes);
   return Object.keys(props)
-    .filter(x => Object.keys(propTypes).indexOf(x) < 0)
+    .filter(x => !propTypesKeys.includes(x))
```
mAiNiNfEcTiOn: Ah nice... That is exactly what it should be, the negation eheh 😉 glad you caught it 👍

PR #231, opened 2017-11-15
Updates to React 16

mAiNiNfEcTiOn:
# What does this PR do:
  * Updates react and its related packages to minimum version of 16.1.0.
  * Replaces `enzyme-adapter-react-15` with `enzyme-adapter-react-16` both in `package.json` and `tests/setupTests.js`. 
  * Updates `jest` to `21.3.0-betaª to get a fix for the rAF warning (see more at: https://github.com/facebook/jest/pull/4568#issuecomment-336562985 )
  * Adds the import of the rAF polyfill, via `raf` package, replacing the existing one.
  * Changes the components to replace the usage of `<noscript />` for `null`.
  * Updates the tests to validate against the `renderTree.html()`'s result and also removes the usage of `React.PropTypes` in favour of `prop-types` one.
  * Updates snapshots. 

  * Diffs
  * Check the styleguide result.


tests/unit/badge/__snapshots__/badge.spec.js.snap
```diff
@@ -53,7 +53,7 @@ exports[`Badge should return content inside container if title not provided 1`]
 </div>
 `;
 
-exports[`Badge should return noscript if children and title not provided 1`] = `<noscript />`;
+exports[`Badge should return noscript if children and title not provided 1`] = `""`;
```
yurist38: You could test `return null` cases without snapshot. Just do this: `expect(component.type()).toEqual(null);`

tests/unit/calendar/calendarWrapper.mock.js
```diff
@@ -39,6 +40,6 @@ export default class CalendarWrapper extends React.Component {
 }
 
 CalendarWrapper.propTypes = {
-  initialDates: React.PropTypes.arrayOf(String),
-  minDate: React.PropTypes.string,
+  initialDates: PropTypes.arrayOf(String),
```
reaktivo: Not part of your changes but this can't be right, _right_?

tests/unit/badge/__snapshots__/badge.spec.js.snap
yurist38: Oh, I see you changed that. So that way should work here as well, right?

tests/unit/calendar/calendarWrapper.mock.js
mAiNiNfEcTiOn: You mean the fact that we have propTypes on a wrapper/mock or the fact that I ordered the propTypes?

tests/unit/badge/__snapshots__/badge.spec.js.snap
mAiNiNfEcTiOn: Added assertion to verify that it returns `null`.

tests/unit/calendar/calendarWrapper.mock.js
reaktivo: `arrayOf(String)` should be `arrayOf(PropTypes.string)`. The first one will basically accept an array of truthy values.

tests/unit/badge/__snapshots__/badge.spec.js.snap
yurist38: Cool, thanks! ...but do we still really need to compare to snapshot? I would suggest to not create snapshots in these cases and just check the `type()`/`html()`. How do you think guys?

tests/unit/calendar/calendarWrapper.mock.js
mAiNiNfEcTiOn: @reaktivo I think the `arrayOf` receives the Constructor/Type. How would you do with a custom class/Component? `PropTypes.MyComponent`?

mAiNiNfEcTiOn: Ahh screw it, I do think this is supported, but anyway... Changing 🙄

mAiNiNfEcTiOn: Now I got it... It's not a PropType checker... just a constructor ... I thought the arrayOf would do (internally) `value instanceof ConstructorThatWasPassedInTheArrayOf`.

tests/unit/badge/__snapshots__/badge.spec.js.snap
mAiNiNfEcTiOn: Strangely, the `type()` did not work... Don't know why, though.

mAiNiNfEcTiOn: Removed

yurist38:
@mAiNiNfEcTiOn should we delete branch?

PR #232, opened 2017-11-15
Refactoring of List component

### LINKED ISSUE #215: Refactor List component to get rid of renderEmptyItems attr
We merged #213 recently but the question (do we really need that prop?) is still open. Most probably we can completely remove it and just do not render empty items for all cases. If you know any case when you need to render empty item(s) within the `List` component please add your comment here. Otherwise, we have to refactor that functionality later on.

yurist38:
# What does this PR do:

  * Removed `renderEmptyItems` from props (fixed #215). Seems like there is no case when we need to render empty items (see the issue for more information).

We just added it some days ago. I used that prop once and it's not live yet. I don't think that anyone else used it recently (if you did that - fix it! 😄 ).

  * Diff


tests/unit/list/__snapshots__/list.spec.js.snap
```diff
@@ -25,7 +25,7 @@ exports[`List #render() should not modify mods 1`] = `
 </ul>
 `;
 
-exports[`List #render() should not render empty items if an appropriate prop is false 1`] = `
+exports[`List #render() should not render empty items 1`] = `
```
leandrooriente: Is this `1` at the end of the message a typo?

mAiNiNfEcTiOn: No... that's the snapshot creator numbering the snapshot check

components/list/list.js
```diff
@@ -26,9 +25,7 @@ function List(props) {
 
   const notEmpty = i => i !== '' && i !== null;
 
-  const itemsToRender = (renderEmptyItems === false) ? items.filter(notEmpty) : items;
-
-  const itemsBlock = itemsToRender.map((item, index) => (
+  const itemsBlock = items.filter(notEmpty).map((item, index) => (
```
maximuk: Why can't we use `filter(Boolean)`?

yurist38: In that case, we will filter out `0` (as Number) values. It's probably the very rare case but still, there is a possibility to get it. And that will be a confusion I think.

maximuk: But we don't have `number` in `PropTypes` for the property

yurist38: That's a good point. 😄  Okay, it makes sense to change it then.

PR #235, opened 2017-11-17
Add footer to sliding panel

### LINKED ISSUE #220: Refactor slidingPanel (add sticky footer)
# Description

Right now we have already re-implemented multiple times sticky footer across all the projects. This functionality should be moved into UI-KIT codebase.

yurist38:
# What does this PR do:
* Fixes #220 
* Added footer (sticky) into the `slidingPanel` component. In a case when content is not fit into the main content block, it (main content block) becomes scrollable (header and footer are sticky).
* I checked all elements with the latest design, they look correct.

screen shot 2017-11-17 at 17 13 57

* Diff

# Acceptance link:
* https://site-aakdzbrgyj.now.sh


themes/_default.yaml
```diff
@@ -505,6 +505,9 @@ sliding-panel:
     flex: "1 1 100%"
     overflow: "hidden"
     overflow-y: "auto"
+  footer:
+    border-top: "2px solid var(--tx-generic-color-secondary-darker)"
+    height: "80px"
```
iwwwi: is it really different per brand?

tests/unit/slidingPanel/__snapshots__/slidingPanel.spec.js.snap
```diff
@@ -102,6 +102,34 @@ exports[`SlidingPanel #render() closes the panel when active prop changes to fal
...
+>
+  <div
+    className="ui-sliding-panel-overlay ui-sliding-panel-overlay_hidden"
+    onClick={[Function]}
+  >
+    <div
+      className="ui-sliding-panel ui-sliding-panel_right"
+      style={
+        Object {
+          "width": "480px",
```
iwwwi: can you pls also add the variations of the sliding panel for small and large version and not use the fixed inline styles?

themes/_default.yaml
asci: I guess people just using this file for defining component variables, not theme variables. A bit of misuse, but there is no better alternative, I'd say

yurist38: Basically, I agree with Atrem, I didn't find the better way to define values.

tests/unit/slidingPanel/__snapshots__/slidingPanel.spec.js.snap
yurist38: Good point. Let's discuss that on Monday how to implement that in better way.

yurist38: So, let's keep it for now and refactor in separate PR. Issue is here -> #240

PR #236, opened 2017-11-18
Update react-styleguide to v6

### LINKED ISSUE #234: Bug: GH-Pages - Show code button does not work
recording 3

any or at least chrome

# Description

clicking "Show code" button-link does not work. In console there is an error:
`TypeError: Cannot read property 'any' of undefined`

# Steps to reproduce

Go to gh-pages, click "show code"

# Expected results

show the correspondent source code

# Actual results

Element is not opening.

yurist38:
# What does this PR do:
* Fixes #234 
* Interface elements are slightly different in the new version.
* Package `eslint-plugin-react` is updated, but after a few broken builds I found out that latest version (7.5.0), released just hours ago, doesn't work well and crashes `npm run lint` script. So fixed version for now.
* Fixed lint errors.

screen shot 2017-11-18 at 19 18 54

* Diff

# Acceptance link:
* [https://site-fiwawlqqnx.now.sh/](https://site-fiwawlqqnx.now.sh/)


yurist38:
So, `eslint-plugin-react` fixed [that issue](https://github.com/yannickcr/eslint-plugin-react/issues/1543) and just released a new version. It works now but shows one warning (Travis does not pass any warnings). I have a doubt how to handle this situation in better way. That warning is about `jsx-no-bind` rule ([read more about it](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md)).
We could either increase `--max-warnings` parameter value in npm script `ci:lint` or disable that rule at all adding it to `.eslintrc`. But ideally, we could refactor the codebase - avoid that warning and get better performance.
The problem here is quite obvious - time resources. So, my suggestion, for now, to disabe that rule temporary and create an issue for refactoring codebase (and enabling rule back). Like if you agree or argue... 😏

mAiNiNfEcTiOn:
From the build I can't see where the warning is thrown. 

Can't we put an eslint-disable-line there?

mAiNiNfEcTiOn:
Ok, so now they're evaluating even if the `.bind()` is done on a 'generated function' ... 👍 

Doing a PR to change that then ;-)

mAiNiNfEcTiOn:
@yurist38 I've created https://github.com/Travix-International/travix-ui-kit/pull/239 to solve this issue 👍

PR #239, opened 2017-11-20
Improves the calendar/views/days.js to not use .bind() according to react/jsx-no-bind lint rule

mAiNiNfEcTiOn:
# What does this PR do:

* Changes the `onSelectDay` method on `calendar.js` to use the `data-date` attribute from the `event's currentTarget` (which is provided from the `calendar/views/days.js` button.

  - Diffs


yurist38:
Great job! Thank you for fixing that quickly! 👍

PR #241, opened 2017-11-22
Updated deploy:accp script to prepend the generated now url with branch name

reaktivo:
# What does this PR do:

  - Updates the `deploy:accp` script to prepend the generated now url with the git branch name.
  - It additionally removes previous deployments under the same branch name in order to prevent you from hitting `now`'s limit on deployments.

  - Run `npm run deploy:accp` and you should notice the generated url is in the format of `https://set-branch-name-on-deploy-accp-rnhcaoqrhp.now.sh/`

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,10 +6,12 @@
   "scripts": {
     "build:watch": "node ./builder/program -w",
     "build": "node ./builder/program",
+    "ci:comment-url": "curl https://api.github.com/repos/$TRAVIS_REPO_SLUG/issues/$TRAVIS_PULL_REQUEST/comments -u \"$TRAVIX_SERVICE_ACCOUNT:$TRAVIX_SERVICE_ACCOUNT_TOKEN\" --data \"{ \\\"body\\\": \\\"A new acceptance URL has been deployed: [$ACCP_URL]($ACCP_URL)\\\" }\"",
     "ci:lint": "eslint --color --max-warnings 0 --quiet '{components,tests,utils,scripts}/**/*.js'",
     "ci:send-coverage": "codecov",
     "ci:test": "TZ=utc jest -c ./tests/jest.config.json --bail --ci --coverage --maxWorkers=2 --no-cache --silent",
-    "deploy:accp": "make prepare-site && now _site --public",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
     "ci:lint": "eslint --color --max-warnings 0 --quiet '{components,tests,utils,scripts}/**/*.js'",
     "ci:send-coverage": "codecov",
     "ci:test": "TZ=utc jest -c ./tests/jest.config.json --bail --ci --coverage --maxWorkers=2 --no-cache --silent",
-    "deploy:accp": "make prepare-site && now _site --public",
+    "deploy:accp": "BRANCH=$(git rev-parse --abbrev-ref HEAD); make prepare-site && now rm -y $BRANCH 2>/dev/null; now _site --public -n $BRANCH",
```
mAiNiNfEcTiOn: Hmmm... will this work on Windows machines? Maybe it would be better to use the CI's env variable for the branch, in this case `TRAVIS_PULL_REQUEST_BRANCH` ? (See more at: https://docs.travis-ci.com/user/environment-variables/ )

reaktivo: We could create a travis specific `now` account and find a way of skip authentication, but as I said this is meant to work locally. I didn't test on Windows, I since I assumed `make` wouldn't work there. Is there anybody here who can verify?

mAiNiNfEcTiOn:
Also if we want to make it in the CI, we just need to have (AFAIK) a login token: https://zeit.co/account/tokens

And then use it with `now login -t <token>` and then use the rest of the commands?

package.json
mAiNiNfEcTiOn: Can you at least use the Travis CI's Environment variable instead of creating the 'BRANCH' one?

reaktivo:
@mAiNiNfEcTiOn Do you have a suggestion on how to actually get the url generated out of Travis CI? Seems like your suggestion implies going into the logs and fetching it out. I'm looking for a way to automate adding a comment on the PR comment thread, but can't find a straightforward way of doing this.

package.json
mAiNiNfEcTiOn: Ahhh forget it. This is for the manual process... Oy oy, my brain was not getting it lol. 👍 Fine by me then :)

mAiNiNfEcTiOn:
Test

mAiNiNfEcTiOn:
@reaktivo I have a way to do so, by using Github API, but I'd rather have a Travix service account doing so...

PR #250, opened 2017-11-27
Add unstyled prop to Price component

yurist38:
# What does this PR do:

> Added `unstyled` prop into `Price` component to have a possibility to render price inline without any styling.

> Review changes and check them on our new awesome acceptance! 😄

### DIFF
--- a/tests/unit/price/price.spec.js
+++ b/tests/unit/price/price.spec.js
@@ -213,7 +213,25 @@ describe('Price', () => {
     });
 
     it('should render unstyled block if an appropriate prop is passed', () => {
-      const wrapper = shallow(<Price unstyled value={50153} />);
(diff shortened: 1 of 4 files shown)


tests/unit/price/price.spec.js
```diff
@@ -211,5 +211,11 @@ describe('Price', () => {
 
       expect(wrapper.html()).toEqual(null);
     });
+
+    it('should render unstyled block if an appropriate prop is passed', () => {
+      const wrapper = shallow(<Price unstyled value={50153} />);
```
asci: can you add more example with decimals?

yurist38: ah, right. Will update

yurist38:
@maximuk you right, we should consider some props for unstyled version. I'm refactoring that.

PR #256, opened 2017-12-04
Implemented Card component

### LINKED ISSUE #255: Card component

EduardTrutsyk:
# What does this PR do:
* implemented Card component
* Closes #255 

    Please, seed diffs.

screenshot at 16-21-39


froskie:
I don't see the checkmark on the acceptance link.

EduardTrutsyk:
@froskie what do you mean?

froskie:
@EduardTrutsyk aren't we supposed to see the check mark on the top right?

spzm:
@froskie click this checkbox:
selection_352

spzm:
Do we have a design to verify?

yurist38:
Is it only for me like this? The icon is different and doesn't fit into triangle...
screen shot 2017-12-05 at 09 15 14

spzm:
+1. In the last acceptance check mark is changed and it doesn't fit anymore.

EduardTrutsyk:
because this icon from default font.

components/card/card.js
```diff
@@ -0,0 +1,99 @@
...
+  /**
+   * Whether to show icon or not
+   */
+  showIcon: PropTypes.bool,
+};
+
+Card.defaultProps = {
+  className: null,
+  dataAttrs: null,
+  hovering: false,
```
froskie: If hovering is just for adding a class name, why simply a mod?

EduardTrutsyk: because our implementation of mods is bad practice, because:
1. difficult to document and validate all variants
2. low performance when using mods
`<Card mods={['hovering']} >...</Card>`
it will create new array each time, when this component rerendered

themes/_default.yaml
```diff
@@ -1,6 +1,8 @@
 generic:
   border:
     radius: &border-radius 8px
+    width: &border-width 2px
```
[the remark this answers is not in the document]
EduardTrutsyk: it's easy way to use the same border size everywhere, instead of hardcode the size.
also easy to customize and refactor.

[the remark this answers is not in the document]
EduardTrutsyk: I'm not sure that good idea to hardcode border width, because sometimes people use 2px sometimes 3px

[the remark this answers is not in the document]
EduardTrutsyk: at this moment, we are using the same border size for the following components(hardcoded value for all of these):
- autocomplete
- button
- card
- checkbox
- datepicker
- dropdown
- input
- radioButton
- toggleButton

I don't know why the variable still does not created.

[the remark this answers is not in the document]
mAiNiNfEcTiOn: @EduardTrutsyk but the question is, are they related? When one changes, should all change?

Not saying that you're wrong, but if you don't know the answers to the questions above (e.g. I don't 😉 😄 ) and you plan to change them, please ask the designers if all those borders are related.

Otherwise, if they're not, you'll have a problem if all use the same variable or custom variables dependent on a main one.

Hardcoding makes sense when the value of the _possible_ variable would not be used/shared amongst components (therefore, as @iwwwi said, not overwritten).

components/card/card.js
asci: `mods` only for custom modifications

components/card/card.js
```diff
@@ -0,0 +1,99 @@
...
+  mods: [],
```
asci: potential object reference issue

components/card/card.scss
```diff
@@ -0,0 +1,48 @@
+.ui-card {
+  border-radius: var(--tx-generic-border-radius);
+  flex: 1 1 auto;
+  position: relative;
+  background: var(--tx-generic-color-blank);
+  border: var(--tx-generic-border-width) var(--tx-generic-border-style) var(--tx-generic-color-blank);
+}
+
+.ui-card_checked.ui-card_show-icon {
+  &:after {
```
asci: please avoid using SCSS features, you can have here simple CSS file which will not be transpiled

components/card/card.scss
```diff
@@ -0,0 +1,48 @@
+.ui-card {
+  border-radius: var(--tx-generic-border-radius);
```
asci: 👍  for using only generic variables

themes/_default.yaml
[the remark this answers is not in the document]
EduardTrutsyk: @mAiNiNfEcTiOn true, but:
1. sometimes raw design has mistakes, and when I notice, that the border size is 3px but in common variable we have 2px, then I will double check with designer, because I need to override it. but maybe it was designers mistake, and they will say: "no, it should be 2px". In this case, we have communication with designers. but when we don't have such common variable, I will just use 3px, like in design.
2. I'm not sure that all guys check the design attentively, or sometimes we don't have design. in this case, developer will use common variables instead of random values.

just my opinion, but if you do not think so, I will delete it.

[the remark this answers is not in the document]
EduardTrutsyk: btw, on design markups for these component we have the same border size, for some of these components we don't have design.

components/card/card.scss
EduardTrutsyk: but why? we have scss in ui-kit, are we going to migrate from SCSS to CSS or css modules?

asci: probably. I'm just saying that you're not using here anything special that could be achieved only with SCSS. Keeping it simple will make any migration in future easier, IMO

components/card/card.js
EduardTrutsyk: actually no, because I'm using `getClassNamesWithMods` helper, and few month ago, I changed this helper and now the mods is immutable.

https://github.com/Travix-International/travix-ui-kit/pull/256/files#diff-2c291c90281191a7a15548daa8de61d2R33

asci: That why I said potential :) just feeling nervous when I see objects in `defaultProps`

asci:
small question: why there is no padding on sides?

EduardTrutsyk:
@asci no need to use padding, because:
1. we have different paddings, depends on situation
2. need to override each time, when you need to add line seporator, separate background or something like that
screenshot at 15-23-04
screenshot at 15-23-30
screenshot at 15-24-29

PR #263, opened 2017-12-13
fixed regex issue in autocomplete component

fope:
# What does this PR do:

    Updated Regex value to handle special symbols.

    please, see the diff


fope:
bad

PR #265, opened 2017-12-14
Fix sliding panel header styles

### LINKED ISSUE #262: Fix SlidingPanelHeader styles in IE

> Internet Explorer 11

# Description

> SlidingPanelHeader is displayed incorrectly
> ie-sliding-panel

# Steps to reproduce

> Run styleguidist in IE

# Expected results

> Buttons at the edges should be centered by vertical

JutIgor:
# What does this PR do:

This PR Closes #262 (SlidingPanelHeader is displayed incorrectly in IE)

Diff


themes/_default.yaml
```diff
@@ -513,11 +513,12 @@ sliding-panel:
       font-size: "32px"
       height: "25px"
       width: "20px"
-      right: "15px"
     back-block:
       font-size: "13px"
       icon:
         font-size: "25px"
+    block:
+      lateral-margin: "15px"
```
yurist38: Is this value different across the themes?

JutIgor: This value (as I remember) I took from `right` variable. The cross was postponed in 15px, and from left side can be `back` arrow. And when `back` arrow is absent, we have wrong aligning, that's why I just move this value to common variable and apply it to both sides.

PR #276, opened 2017-12-19
Carousel - fixed transition for Safary 9

maximuk:
# What does this PR do:
* fixed transition for Safary 9
* diff


yurist38:
We have a problem with loading pics in carousel because in the `.md` file they defined via `http` protocol but we mostly use `https` (including `now` deployments). So I think we should change `.md` to use `//domain` instead of `http://domain` to support all protocols. @maximuk could you please test it and change...

PR #277, opened 2017-12-19
Remove otherProps from Tab/Tabs components

juliamaksimchik:
# What does this PR do:

* Removed usage of other props from Tab/Tabs component

diff

### DIFF
--- a/components/tabs/tabs.js
+++ b/components/tabs/tabs.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 
 import Tab from './tab';
-import { getClassNamesWithMods, getDataAttributes, ejectOtherProps } from '../_helpers';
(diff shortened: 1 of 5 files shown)


components/tabs/tabs.js
```diff
@@ -153,7 +150,6 @@ class Tabs extends Component {
 
 Tabs.defaultProps = {
   activeTab: '0',
-  open: false,
```
juliamaksimchik: I've noticed that this prop is not used in the component, thus, I'm deleting it. Please correct me if I am wrong

iwwwi:
@juliamaksimchik can you pls explore the possibility to increase the accessibility with the Tabs? Here are some examples how it can be achieved: 
http://accessibility.athena-ict.com/aria/examples/tabpanel2.shtml

PR #278, opened 2017-12-20
List component fixes

annavolkovskaya:
# What does this PR do:

    List component fixes

    diffs

### DIFF
--- a/components/list/list.js
+++ b/components/list/list.js
@@ -1,7 +1,8 @@
 // Imports
 import PropTypes from 'prop-types';
 import React from 'react';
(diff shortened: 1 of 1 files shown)


maximuk:
Could you please also remove `hideBullets` prop from https://github.com/Travix-International/travix-ui-kit/blob/list-component-fixes/tests/unit/list/list.spec.js#L40

components/list/list.js
```diff
@@ -11,6 +12,8 @@ function List(props) {
     align,
     hideBullets,
     items,
+    className,
```
sleepwalky: Should it be sorted there?

PR #279, opened 2017-12-20
Collapse fix

maximuk:
# What does this PR do:
* fixed styles for firefox and safari 9 for collapse and collapseItem components
* removed otherProps and added dataAttrs to collapse and collapseItem components
* added className prop to collapse and collapseItem components
* diff
#### safari before
selection_099
#### safari after
selection_098
#### firefox before
selection_100

### DIFF
--- a/components/collapse/collapseItem.js
+++ b/components/collapse/collapseItem.js
@@ -1,5 +1,6 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
+import classnames from 'classnames';
 
 import { getClassNamesWithMods, getDataAttributes } from '../_helpers';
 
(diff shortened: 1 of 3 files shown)


components/collapse/collapseItem.js
```diff
@@ -17,22 +18,29 @@ class CollapseItem extends Component {
...
       id,
       isActive,
       title,
       children,
-      onClick, // eslint-disable-line no-unused-vars
       iconPosition,
-      labelProps = {},
-      ...otherProps
+      dataAttrs,
+      labelDataAttrs,
```
iwwwi: are we sure that here we pass only dataAttrs for the label?

maximuk: Yes, we are. We pass this prop to `getDataAttributes` function and the function returns only data attributes

iwwwi: 👍

PR #280, opened 2017-12-20
Calendar component fixes for IE and Edge

juliamaksimchik:
# What does this PR do:
 

-  Fixed [flex-basis bug](https://github.com/philipwalton/flexbugs#7-flex-basis-doesnt-account-for-box-sizingborder-box) in IE (workaround with width). 

-  Set `overflow:visible` prop for `button` to make pseudo-elements (before, after) visible in IE

-  Added a fix for IE and Edge to solve [border radius bug ](https://long2know.com/2015/09/ie11-rounded-corner-bug/) (white 1px line in calendar navigation)

-  `document.getElementById('output_2').value` is not working in IE, thus, replaces it with `document.getElementById('output_2').innerText`

    diff

Before in IE11:
ie-11-before

After in IE11:
ie-11-after

### DIFF
--- a/components/calendar/calendar.js
+++ b/components/calendar/calendar.js
@@ -247,7 +247,6 @@ export default class Calendar extends Component {
 }
 
 Calendar.defaultProps = {
-  className: '',
   multiplemode: false,
   selectionType: 'normal',
 };
--- a/components/calendar/calendar.js
+++ b/components/calendar/calendar.js
@@ -1,5 +1,6 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
+import classnames from 'classnames';
 import { getClassNamesWithMods, getDataAttributes, normalizeDate } from '../_helpers';
 import DaysPanel from './panels/days';
 import calendarConstants from './constants/calendar';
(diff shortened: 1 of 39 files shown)


AlleeX:
:muscle:

components/calendar/calendar.js
```diff
@@ -213,15 +214,19 @@ export default class Calendar extends Component {
   }
 
   render() {
-    const { dataAttrs = {}, isDaySelectableFn, locale, navButtons, selectionType } = this.props;
+    const { dataAttrs = {}, isDaySelectableFn, locale, navButtons, selectionType, className } = this.props;
```
maximuk: default value for `dataAttrs` is not necessary, because we check it in `getDataAttributes` function

juliamaksimchik: Fixed

components/calendar/calendar.scss
```diff
@@ -26,13 +26,19 @@
...
-  border-radius: var(--tx-calendar-nav-border-radius);
   display: flex;
   flex: 0 0 var(--tx-calendar-nav-height);
   flex-flow: row nowrap;
   width: 100%;
 }
 
+.ui-calendar-days__head {
+  display: flex;
+  flex-direction: column;
```
AntonTsymuk: please check sorting

components/calendar/calendar.scss
```diff
@@ -43,7 +49,8 @@
   box-shadow: 0 2px var(--tx-calendar-nav-button-box-shadow-color);
   cursor: pointer;
   color: var(--tx-calendar-nav-button-color);
-  flex: 0 0 var(--tx-calendar-nav-button-width);
+  flex: 0 0 auto;
+  width: var(--tx-calendar-nav-button-width);
   font-weight: var(--tx-calendar-nav-button-font-weight);
```
svetlanamorozko: Please, check sorting

themes/_default.yaml
```diff
@@ -201,9 +201,10 @@ button:
       decoration: "underline"
 calendar:
   height: 300px
+  head:
+    border-radius: "5px 5px 0 0"
```
iwwwi: before moving this variable here pls confirm that it is not overwritten in the travix-themes

juliamaksimchik: checked! It's not overwritten

iwwwi: if it is not used at all pls remove the variable and add it directly to the sass file as value... :)

PR #282, opened 2017-12-20
Cfe 17504 button

svetlanamorozko:
# What does this PR do:
Fix button styles for IE11

```
diff
```

### Before in IE11:
before1
before2

### After in IE11:
after1
after2

### DIFF
--- a/components/autoComplete/autoComplete.scss
+++ b/components/autoComplete/autoComplete.scss
@@ -12,6 +12,7 @@
   margin-top: -2px;
   max-height: 300px;
   overflow-x: overlay;
+  overflow-y: auto;
   position: absolute;
   right: 0;
   top: 100%;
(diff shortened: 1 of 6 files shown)


components/button/button.js
```diff
@@ -12,19 +12,18 @@ function Button(props) {
...
+    size,
     type,
     variation,
-    disabled,
-    dataAttrs = {},
   } = props;
   const restProps = getDataAttributes(dataAttrs);
   const mods = props.mods ? props.mods.slice() : [];
 
-  const otherProps = ejectOtherProps(props, Button.propTypes);
```
yurist38: Why are you removing `otherProps`?

iwwwi: It is a decision to not have `otherProps` anymore passed in ui-kit.. This opens vulnerability to pass properties that are not suitable for the DOM and also does allow a lot of freedom. The decision is, in future to pass only known properties - that are publicly documented in the ui-kit. If there is a need to extend the current list of properties - please make sure that proper PRs are added.

iwwwi:
@AlleeX Can you pls provide the example to @svetlanamorozko so she can update the PR accordingly?

PR #284, opened 2017-12-21
Add className property Checkbox and RadioButton components

AntonTsymuk:
# What does this PR do:

-     add className property Checkbox and RadioButton components

### DIFF
--- a/components/checkbox/checkbox.js
+++ b/components/checkbox/checkbox.js
@@ -21,7 +21,7 @@ function Checkbox(props) {
   const dataAttributes = getDataAttributes(dataAttrs);
   const mods = props.mods ? props.mods.slice() : [];
   disabled && mods.push('is-disabled');
-  const classNames = classnames(className, getClassNamesWithMods('ui-checkbox', mods));
+  const classNames = classnames(getClassNamesWithMods('ui-checkbox', mods), className);
 
   return (
     <label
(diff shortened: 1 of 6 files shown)


iwwwi:
can you pls add additionally `id` prop to checkbox component? Tnx

AntonTsymuk:
@iwwwi there is already `name` prop which is used as id for input. should i rename it to `id`?

components/radioButton/radioButton.js
```diff
@@ -11,23 +12,24 @@ function RadioButton(props) {
...
-  const restProps = getDataAttributes(dataAttrs);
+  const dataAttributes = getDataAttributes(dataAttrs);
   const mods = props.mods ? props.mods.slice() : [];
 
   if (disabled) {
     mods.push('disabled');
   }
 
-  const className = getClassNamesWithMods('ui-radio', mods);
+  const classNames = classnames(className, getClassNamesWithMods('ui-radio', mods));
```
asci: should you give priority to className instead of ui-kit classes?

components/checkbox/checkbox.js
```diff
@@ -18,12 +21,12 @@ function Checkbox(props) {
   const dataAttributes = getDataAttributes(dataAttrs);
   const mods = props.mods ? props.mods.slice() : [];
   disabled && mods.push('is-disabled');
-  const className = getClassNamesWithMods('ui-checkbox', mods);
+  const classNames = classnames(className, getClassNamesWithMods('ui-checkbox', mods));
```
asci: Maybe put `className` as the last argument?

PR #285, opened 2017-12-21
Fixed input component

annavolkovskaya:
# What does this PR do:

    Fixed input component

   Diffs

# Note

 Decision to leave border the same color on hover for disabled input and give disabled input 0.5 opacity was made by Nicolas.

### DIFF
--- a/components/input/input.js
+++ b/components/input/input.js
@@ -1,7 +1,8 @@
 import PropTypes from 'prop-types';
+import classnames from 'classnames';
 import React, { Component } from 'react';
 
-import { getClassNamesWithMods, getDataAttributes, ejectOtherProps } from '../_helpers';
+import { getClassNamesWithMods, getDataAttributes } from '../_helpers';
 
 /**
  * General Input component.
@@ -68,6 +69,7 @@ class Input extends Component {
 
   render() {
     const {
+      className,
       dataAttrs = {},
       disabled,
       multiline,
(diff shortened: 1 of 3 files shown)


components/input/input.js
```diff
@@ -76,21 +78,21 @@ class Input extends Component {
       value,
     } = this.props;
     const mods = this.props.mods ? this.props.mods.slice() : [];
-
-    const otherProps = ejectOtherProps(this.props, Input.propTypes);
```
yurist38: What is the reason for getting rid of `otherProps`?

components/input/input.scss
```diff
@@ -53,6 +53,14 @@
     border-color: var(--tx-input-border-hover-color);
   }
 
+  &.ui-input_disabled {
+    opacity: 0.5;
```
yurist38: I don't think that applying opacity is a good way of showing disabled state. I rather prefer to play with colors instead.

annavolkovskaya: As I have written in PR's note the decision to make such style changes was made by designer Nicolas.

components/input/input.js
annavolkovskaya: Because we have such a task. We need to define what props are these 'otherProps'

components/input/input.scss
yurist38: I think that we need to discuss that. And if everyone agrees - apply this styling for all disabling states to avoid inconsistency. What do you guys think about semi-transparency for the disabled state? My concern is looking will depend on background, I think that it's not good. We have a lot of specific hues for each theme, better to use one of them for disabled state.

components/input/input.js
yurist38: That's why I ask everyone to put more details into their PR's description. To reduce this kind of questions.

components/input/input.scss
iwwwi: @yurist38 when new designs were introduced for the input elements the disabled state was missed out. @annavolkovskaya was talking to the designers and this is for now the solution... therefore until we have other design introduced we will use this approach ;)

PR #286, opened 2017-12-21
Fix theme for messagebox

sleepwalky:
# What does this PR do:

- Add className prop

- Fix success color css-var

diff

### DIFF
--- a/components/messageBox/messageBox.js
+++ b/components/messageBox/messageBox.js
@@ -1,9 +1,19 @@
 import PropTypes from 'prop-types';
 import React from 'react';
+import classnames from 'classnames';
 import { getClassNamesWithMods, getDataAttributes } from '../_helpers';
 
-const MessageBox = ({ children, dataAttrs, icon, mods, title, type }) => {
-  const className = getClassNamesWithMods('ui-messageBox', [...mods, type]);
+const MessageBox = (props) => {
+  const {
+    children,
+    className,
+    dataAttrs,
+    icon,
+    mods,
+    title,
+    type,
+  } = props;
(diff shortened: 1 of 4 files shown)


components/messageBox/messageBox.js
```diff
@@ -1,9 +1,19 @@
 import PropTypes from 'prop-types';
 import React from 'react';
+import classnames from 'classnames';
 import { getClassNamesWithMods, getDataAttributes } from '../_helpers';
 
-const MessageBox = ({ children, dataAttrs, icon, mods, title, type }) => {
-  const className = getClassNamesWithMods('ui-messageBox', [...mods, type]);
+const MessageBox = (props) => {
+  const {
+    children,
```
juliamaksimchik: Pls, sort it

sleepwalky: But... isn't it sorted?

themes/_default.yaml
```diff
@@ -400,7 +400,7 @@ messagebox:
   background-color:
     info: "var(--tx-generic-color-secondary-light)"
     error: "var(--tx-generic-color-negative-light)"
-    success: "var(--tx-generic-color-positive-lighter)"
+    success: "var(--tx-generic-color-positive-light)"
```
iwwwi: is this according to our designs?

sleepwalky: @iwwwi 
well, I've just fixed typo

maximuk: We have this color in `travix-themes` and you shouldn't change the variable. I think you can just set any color for `positive-lighter` in `_default.yaml` file

sleepwalky: @iwwwi @maximuk 
I've rolled it back, guys!

PR #289, opened 2017-12-22
fix styling issue with custom marker; add className and dataAttr properties

AntonTsymuk:
# What does this PR do:

- fix styling issue with custom marker

googlemap travix styleguide

- add className and dataAttr properties

### DIFF
--- a/components/googleMap/googleMap.js
+++ b/components/googleMap/googleMap.js
@@ -3,6 +3,8 @@ import React, { Component } from 'react';
 import ReactDOM from 'react-dom';
 import classnames from 'classnames';
 
+import { getDataAttributes } from '../_helpers';
+
 /**
  * Initializes custom marker constructor
  *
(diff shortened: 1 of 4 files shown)


components/googleMap/googleMap.js
```diff
@@ -231,10 +232,10 @@ export default class GoogleMap extends Component {
   }
 
   render() {
-    const { width, height } = this.props;
+    const { width, height, className } = this.props;
```
iwwwi: can you pls sort the props alphabetically :)

components/googleMap/googleMap.js
```diff
@@ -247,6 +248,10 @@ GoogleMap.propTypes = {
    * Map center coordinates.
    */
   center: PropTypes.object,
+  /**
+   * Attribute used to set specific classes
+   */
+  className: PropTypes.string,
```
iwwwi: can you pls also add the dataAttr prop?

PR #290, opened 2017-12-22
Add className property to Spinner

svetlanamorozko:
# What does this PR do:
Add className property

diff

### DIFF
--- a/components/spinner/spinner.js
+++ b/components/spinner/spinner.js
@@ -1,4 +1,5 @@
 // Imports
+import classnames from 'classnames';
 import PropTypes from 'prop-types';
 import React from 'react';
 import { getClassNamesWithMods } from '../_helpers';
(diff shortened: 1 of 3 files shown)


components/spinner/spinner.js
```diff
@@ -1,22 +1,19 @@
 // Imports
```
juliamaksimchik: Do we need this comment?

components/spinner/spinner.js
```diff
@@ -1,22 +1,19 @@
...
 import React from 'react';
-import { getClassNamesWithMods } from '../_helpers';
 
 /**
  * General Spinner component. Use when you need spinner
  */
 function Spinner(props) {
-  const { size } = props;
-  const mods = props.mods ? props.mods.slice() : [];
+  const { className, size } = props;
```
juliamaksimchik: Check with Ivana if we should remove props

yurist38: Yeah, I guess we have decided that mods are kinda deprecated but still in place. They should be removed later.

components/spinner/spinner.scss
```diff
@@ -10,8 +10,8 @@
 }
 
 .ui-spinner_size_s {
-  height: 37px;
-  width: 37px;
+  height: 38px;
```
iwwwi: what is the reason to change this value?

PR #292, opened 2017-12-22
Fixed price component

annavolkovskaya:
# What does this PR do:

    Added className to price component

    diffs

### DIFF
--- a/components/price/price.js
+++ b/components/price/price.js
@@ -171,7 +171,7 @@ Price.propTypes = {
   additionalText: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
 
   /**
-   * Classname for input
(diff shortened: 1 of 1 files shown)


asci:
Would be nice to also add with `className` deprecation warning related to usage of mods. Otherwise, it looks a bit inconsistent.

components/price/price.js
```diff
@@ -168,6 +170,11 @@ Price.propTypes = {
    */
   additionalText: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
 
+  /**
+   * Classname for input
```
iwwwi: can you pls update the docs?
ex: `Custom classname` will be ok :)

iwwwi: can you pls update the docs?
ex: `Custom classname` will be ok :)

PR #294, opened 2017-12-26
Add className and dataAttrs to Modal component's props 

sleepwalky:
# What does this PR do:

 Adds className and dataAttrs to props
 Fix scroll in IE

Before:
image

image

After:
image

image

diff


sleepwalky:
@juliamaksimchik 
done!

components/modal/modal.js
```diff
@@ -228,6 +238,13 @@ Modal.propTypes = {
    * Determine whether to close the modal dialog when clicked on overlay.
    */
   closeOnOverlayClick: PropTypes.bool,
+  /**
+   * Data attribute. You can use it to set up any custom data-* attribute.
+   */
+  dataAttrs: PropTypes.oneOfType([
+    PropTypes.bool,
```
yurist38: For me it's not really clear why we support boolean as a proptype for dataAttrs. But it's already implemented this way, probably there is a reason for that. Just curious... 🤔

components/modal/modal.scss
```diff
@@ -133,6 +131,7 @@
 .ui-modal__content {
   -webkit-overflow-scrolling: touch;
   background-color: var(--tx-modal-content-background-color);
+  max-height: calc(54vh);
```
yurist38: Why do you need `calc` here? Just `54vh` is enough. Another question is why it's exactly 54?

yurist38: It's not really critical, but It's would be nice to change the layout for `.ui-modal__container` and make it `display: flex`. By that, we can specify `max-height` for that main container only and children will be settled automatically. Right now I don't see any option than set up some fixed value (same as you did, but without `calc` please 😄 )

sleepwalky: @yurist38 
mmm, just doing some explorations because max-height is not working as expected in IE
https://github.com/philipwalton/flexbugs/issues/216

yurist38: Okay, I see. I'm fine with this solution for now once you have removed `calc()`... 😄

froskie:
Closed due to outdated code and important part moved to https://github.com/Travix-International/travix-ui-kit/pull/365

PR #299, opened 2017-12-27
Rename

ibsukru:
# What does this PR do:
Fix naming in the docs
diff


components/rating/rating.js
```diff
@@ -5,9 +5,11 @@ import React from 'react';
 import { getDataAttributes } from '../_helpers';
 
 /**
- * Ratint component
+ * Rating component
```
iwwwi: can you also pls add the possibility to pass className?

PR #301, opened 2017-12-28
[POC] usage of travix-themes

maximuk:
# What does this PR do:
* added possibility to change theme
* added possibility to use custom theme package

# Screenshots:
* selection_108
* selection_107
* selection_106
* selection_105
* selection_104

# TODO:
* update documentation
* fix unit tests

# Let's discuss it


components/themes.js
```diff
@@ -0,0 +1,66 @@
...
+    document.head.appendChild(this.themeJS);
+
+    this.polyfilljs = document.createElement('script');
+    this.polyfilljs.src = '/css-vars-polyfill.js';
+    this.polyfilljs.async = false;
+    document.head.appendChild(this.polyfilljs);
+  }
+
+  componentWillMount() {
+    this.onChange(this.state);
```
yurist38: I guess here React might throw a warning that you use `setState` on an unmounted component. I would suggest first to move applying css functionality from `onChange` into separate method (same as `applyCssPolyfill`). Then in `componentWillMount` we can just call them 2 directly without setting state. So could be something like this:
```
onChange = (option) => {
  this.setState({ value: option.value });
  this.applyThemeCss();
  this.applyCssPolyfill();
}

componentWillMount() {
  this.applyThemeCss();
  this.applyCssPolyfill();
}
```
Or even I guess we could also move apply theme js into the separate method as well.

builder/builder.js
```diff
@@ -37,41 +38,48 @@ module.exports = (options) => {
...
   } = options;
 
-  const themeFiles = [defaultThemeYamlPath].concat(themeFile).filter(Boolean);
+  const bundles = [{
+    themeFiles: [defaultThemeYamlPath],
+    output: path.join(defaultOutputThemeDir, 'default'),
+  }];
+  getAvailableSets().forEach((themeOptions) => {
+    bundles.push({
+      themeFiles: [defaultThemeYamlPath].concat(getThemeFiles(themeOptions)),
```
yurist38: I would recommend simplifying a bit here preparing data before calling the `push`. It will increase readability.
```
getAvailableSets().forEach((themeOptions) => {
  const themeFiles = [defaultThemeYamlPath].concat(getThemeFiles(themeOptions));
  const output = path.join(defaultOutputThemeDir, `${themeOptions.brand}-${themeOptions.affiliate}`);
  bundles.push({themeFiles, output});
});
```

builder/builder.js
```diff
@@ -37,41 +38,48 @@ module.exports = (options) => {
...
+      output: path.join(defaultOutputThemeDir, `${themeOptions.brand}-${themeOptions.affiliate}`),
+    });
+  });
 
   const builder = themeBuilder({
```
yurist38: I suggest renaming it to the `builderCSS`

builder/builder.js
```diff
@@ -37,41 +38,48 @@ module.exports = (options) => {
...
+        outputFile(`${output}.js`, `window.TravixTheme = ${JSON.stringify(result)};`);
+      });
     });
   }
 
-  return Promise.all([
-    buildThemeCSS(builder, themeFiles, output),
-    buildThemeJS(builderJS, themeFiles, outputThemeJs),
-  ]).then(() => runWebpackAndCopyFilesToFinalDestination({
+  return Promise.all(bundles.reduce((result, { themeFiles, output }) => {
```
yurist38: And again here, could be simplified. It's fine for POC and we can read it of course, but... It takes time, a lot of nested stuff. Let's try to go more KISS way and write a bit simpler. Assume that your code will be maintained by babies 😄

yurist38:
Would be great to have some space between the text and dropdown...

screen shot 2017-12-28 at 16 15 48

package.json
```diff
@@ -91,6 +90,7 @@
     "sass-loader": "~6.0.4",
     "style-loader": "~0.18.2",
     "theme-builder": "~0.5.0",
+    "travix-themes": "git+ssh://git@bitbucket.org/xivart/travix-themes.git#ui-kit-plus-travix-themes-poc",
```
iwwwi: Because this is pointing to private repo.. TravisCI wont be able to clone this repo.. I would suggest you to copy the file structure from the repo initially in this repository ... and not to have to point here :)

builder/builder.js
```diff
@@ -37,41 +38,48 @@ module.exports = (options) => {
...
 
-  const themeFiles = [defaultThemeYamlPath].concat(themeFile).filter(Boolean);
+  const bundles = [{
+    themeFiles: [defaultThemeYamlPath],
+    output: path.join(defaultOutputThemeDir, 'default'),
+  }];
+  getAvailableSets().forEach((themeOptions) => {
+    bundles.push({
+      themeFiles: [defaultThemeYamlPath].concat(getThemeFiles(themeOptions)),
+      output: path.join(defaultOutputThemeDir, `${themeOptions.brand}-${themeOptions.affiliate}`),
```
iwwwi: this will result with breaking change .. just keep in mind that in `rwd` we have usage of the theme.css

maximuk:
To use custom package run the following commands:
1. `npm run build:watch -- -t <path>`
2. `npm run styleguide-server <path>`

It's useful for development, because we shouldn't link package with themes. By default we use `travix-themes` package

yurist38:
@maximuk I think README file also needs to be adjusted according to these changes. By the way, you did a great job! Thank you!

froskie:
This was closed to wait the update in the mono repo and then proceed with this.

PR #302, opened 2018-01-04
Fix dropdown styles

svetlanamorozko:
# What does this PR do:
Fix dropdown styles according the new design

diff

### DIFF
--- a/components/dropDown/dropDown.scss
+++ b/components/dropDown/dropDown.scss
@@ -36,8 +36,11 @@ $select-arrow-color: var(--tx-dropdown-arrow-color);
   }
 
   .Select-control {
+    display: flex;
+    justify-content: space-between;
+
     .Select-placeholder {
-      color: var(--tx-generic-color-secondary-darker);
+      color: var(--tx-generic-color-text-light);
     }
   }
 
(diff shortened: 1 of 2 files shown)


components/dropDown/dropDown.scss
```diff
@@ -23,6 +23,10 @@ $select-arrow-color: var(--tx-dropdown-arrow-color);
     }
   }
 
+  &.is-focused:not(.is-open) .Select-control {
+    border-color: $override-select-input-border-color;
```
yurist38: Should we maybe use CSS variable instead, same as in other places?

themes/_default.yaml
```diff
@@ -321,13 +321,14 @@ dropdown:
     box-shadow: "0 0 5px 0 rgba(var(--tx-generic-color-shadow-rgb), .18)"
   select-arrow-zone:
     border-color-left: "var(--tx-generic-color-secondary-darker)"
-    color: "var(--tx-generic-color-active)"
+    border-style: "dotted"
```
iwwwi: is this different per brand? if not, pls add this directly in the css without creating a variable.

PR #308, opened 2018-01-11
Fixed input properties

annavolkovskaya:
# What does this PR do:

    fixed input props

 diffs


components/input/input.js
```diff
@@ -92,13 +101,28 @@ class Input extends Component {
       <div className="ui-input-container">
         <Element
           {...getDataAttributes(dataAttrs)}
+          aria-activedescendant={this.props['aria-activedescendant']}
```
yurist38: What do you guys think about adding a helper passing all `aria-*` attributes whitelisted in `propTypes`? It could look like this:
```
function getAriaAttributes(props, propTypes) {
  return Object.keys(props)
    .filter(
      x => x.toLowerCase().startsWith('aria-') &&
      Object.keys(propTypes).includes(x.toLowerCase())
    )
    .reduce((prev, item) => ({ ...prev, [item]: props[item] }), {});
}
```
We could use it in the same way as `getDataAttributes` is used (instead of writing here a long list of available `aria`s). In this case we just need to define a `propTypes` for aria attributes to whitelist them.

annavolkovskaya: If I understood you correctly, then after the last changes not sure it's better to do it this way

yurist38: yeah, last changes make it a bit harder to implement... But it's doable still. I'm fine with both (but prefer a helper way). Just don't like repeating stuff which can be automated. I think it's our (as developers) direct responsibility to make code short and beautiful.

yurist38: Let's wait for others opinions.

AlleeX: My opinion.. using otherProps for the simple components like 'input', 'button' and etc this is best practice. Because in future, step by step, you will add all native tag API as properties or use some helpers, like @yurist38  suggested. 
Is it really for us the using otherProps - taboo?

сс: @iwwwi @froskie @maximuk @EduardTrutsyk

yurist38: There is another option which I suggested at the first meeting (and it is still relevant for me): we could use this package https://www.npmjs.com/package/html-element-attributes to get the list of all possible attributes for the specific element and pass only listed ones.

annavolkovskaya: So what is the outcome? I like the variant which was suggested by @yurist38 , but need approvals from the others

AlleeX: For me it's fine. Let's use @yurist38 approach ;)

yurist38: I've had a look at that package again, how it works. First of all, it doesn't support `aria` attributes. Another thing is it makes `http(s)` request every time, that will badly affect the performance. So I rather avoid to use it knowing that now.

PR #309, opened 2018-01-12
warning of usage of deprecated "mods" prop

AntonTsymuk:
# What does this PR do:

    Log warning message of usage of deprecated "mods" prop


components/_helpers.js
```diff
@@ -88,11 +88,26 @@ function ejectOtherProps(props, propTypes) {
...
+ * Warn about usage of deprecated prop
+ *
+ * @method warnAboutDeprecatedProp
+ * @param {Any} propValue value of deprecated prop.
+ * @param {String} oldPropName deprecated prop name.
+ * @param {String} newPropName new prop name.
+ */
+function warnAboutDeprecatedProp(propValue, oldPropName, newPropName) {
+  if (propValue !== undefined) {
+    console.warn(`[DEPRECATED] the property "${oldPropName}" has been deprecated, use "${newPropName}" instead`);
```
yurist38: If I understand correctly there is no possibility to switch warnings off on production, right?

AntonTsymuk: as i can see at the moment we do not distribute any development bundles. I am not sure if we should introduce it

PR #311, opened 2018-01-13
Update react-styleguidist

### LINKED ISSUE #310: Preview crashes when name of attribute contain '-' and it is boolean (true)

> Chrome (most probably others also)

# Description

> I found that preview of any component (in styleguidist) crashes when the component has attribute containing dash (`-`) character and having a no value. For example: `<Input aria-haspopup />`
screen shot 2018-01-13 at 02 10 28
> It works fine if the value of the attribute is provided or there is no dash character in the attribute name.

# Steps to reproduce

> Go to the website https://travix-international.github.io/travix-ui-kit/#button
> Open code for any component and add attribute 'foo-bar' without any value

# Expected results

> This kind of attributes should be valid and passed further.

yurist38:
# What does this PR do:

> Closes #310 

> Open issue (#310), read it and try to reproduce the bug. Check in different browsers.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -185,20 +185,12 @@
       "dev": true
     },
     "acorn-dynamic-import": {
-      "version": "2.0.2",
-      "resolved": "https://registry.npmjs.org/acorn-dynamic-import/-/acorn-dynamic-import-2.0.2.tgz",
-      "integrity": "sha1-x1K9IQvvZ5UBtsbLf8hPj0cVjMQ=",
+      "version": "3.0.0",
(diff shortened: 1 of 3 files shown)


yurist38:
I found that latest version of `react-styleguidist` has known issue with IE11. So as you can see the acceptance link doesn't work on IE11 unfortunately. Here is a link to the issue https://github.com/styleguidist/react-styleguidist/issues/764
I will try to investigate and fix it. If anyone (especially who has windows installed) can help with fixing - it will be very appreciated! 😏

yurist38:
Fixed now.

leandrooriente:
Hey @yurist38 can you fix the conflict? 

And why we have a lot of merges from master to ours PRs? Would not be better to rebase it?

PR #312, opened 2018-01-16
Extend spinner using children

RumataDeEstor:
# What does this PR do:

    This PR extends Spinner component to be able to wrap children into the loading section.

 components/spinner/spinner.js

### DIFF
--- a/components/spinner/spinner.scss
+++ b/components/spinner/spinner.scss
@@ -125,7 +125,7 @@
   color: var(--tx-generic-color-primary);
   font-family: var(--tx-generic-font-primary-font-family);
   font-size: 19px;
-  font-weight: bold;
+  font-weight: var(--tx-generic-font-primary-weight-semibold);
   margin-left: 8px;
 }
 
--- a/components/spinner/spinner.scss
+++ b/components/spinner/spinner.scss
@@ -123,7 +123,7 @@
 
 .ui-spinner__loading-text {
   color: var(--tx-generic-color-primary);
-  font-family: var(--tx-generic-font-primary-font-family);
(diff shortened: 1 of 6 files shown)


components/spinner/spinner.scss
```diff
@@ -110,3 +110,51 @@
...
+  .ui-spinner__content:before {
+    background: var(--tx-generic-color-secondary-light);
+    bottom: 0;
+    content: "";
+    left: 0;
+    position: absolute;
+    right: 0;
+    top: 0;
+    transition: all 0.3s;
+    z-index: var(--tx-spinner-content-z-index);
```
froskie: The good ol' logic of applying variables: is thing changing per brand? If not, keep on the component.

RumataDeEstor: Fixed :)

RumataDeEstor: @froskie The prop 'loading' needed when you use the component as a wrapper. It's 'loading' by default, but then when you finish loading your stuff inside the container, you can change it to hide spinner itself and show your content inside the wrapper. So that you don't need to implement it each time. When using component as a normal spinner without children, you won't need to change the prop.

yurist38:
Did you try to fix the small (s) state behavior in IE11? I see that it's moving a bit horizontally during rotation, on production also.

maximuk:
@yurist38 I guess it will be fixed by https://github.com/Travix-International/travix-ui-kit/pull/304

asci:
Have you considered to create another component with name `Loader` which will have this ability to have children and text? In my opinion, this change will lead to misuse of spinner and will make it unnecessary bigger

PR #313, opened 2018-01-18
Add selection block component

AlleeX:
# What does this PR do:

    Add selection block component

selection_121

    Diff


iwwwi:
screen shot 2018-01-19 at 16 46 13

got this while checking in IE 11. pls check why we do not see the borders :)

AlleeX:
@iwwwi regarding IE issue. This is only the demo issue. The same issue we have in production. Because we use css variables in MD files. I'll remove css variables from my demo files.

Issue #314, opened 2018-01-18
Update Modal design to fit mock ups


spzm:
# Browsers and versions affected

Add small size for modal modal. Update current big size to fit design.

PR #321, opened 2018-01-31
Update modal design

spzm:
# What does this PR do:
  
     - The design of modal was updated regarding issue #314.
     - Introduce small and big (by default) states of modal regarding design.
     - ModalContent component introduced to unify content body inside modal window.

    - Test modal window with styleguide.
    - Check backward compatibility according current usage of modal.


components/index.scss
```diff
@@ -17,6 +17,7 @@
 @import "./loadingOverlay/loadingOverlay.css";
 @import "./messageBox/messageBox.scss";
 @import "./modal/modal.scss";
+@import "./modal/modalContent.scss";
```
AlleeX: only css! ;)

components/modal/modalContent.js
```diff
@@ -0,0 +1,39 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+
```
AlleeX: extra line

themes/_default.yaml
```diff
@@ -434,6 +434,10 @@ modal:
       color: "var(--tx-generic-color-active)"
     icon:
       color: "var(--tx-generic-color-secondary-darker)"
+  modal-content:
+    background:
+      color: "var(--tx-generic-color-blank)"
+    border-radius: 8px;
```
AlleeX: are we really have dependency from other themes?

EduardTrutsyk:
why the modal closing without any animation?

components/modal/modal.js
```diff
@@ -256,6 +258,10 @@ Modal.propTypes = {
    * Determine whether a overlay is visible or not
    */
   overlay: PropTypes.bool,
+  /**
+   * Price's size
```
iwwwi: Modal size

spzm:
In master I see same close-without-animation behavior. Also, the last modal has close animation. I'm going to check is it possible to fix with this PR.

components/modal/modalContent.css
```diff
@@ -0,0 +1,38 @@
+.ui-modal-content-block {
+  background-color: var(--tx-generic-color-blank);
+  border-radius: 8px;
+}
+
+.ui-modal-content-block__title {
+  font-weight: 600;
```
maximuk: please sort and use variable for `font-weight` (also in other places)

components/modal/modalContent.css
```diff
@@ -0,0 +1,38 @@
...
+  border-bottom: 2px solid var(--tx-modal-content-background-color);
+  padding: 24px;
+}
+
+.ui-modal-content-block__container {
+  border-radius: 8px;
```
maximuk: we have variable for `border-radius` - `--tx-generic-border-radius` (also should be fixed in other places)

components/modal/modalContent.js
```diff
@@ -0,0 +1,38 @@
...
+ModalContent.propTypes = {
+  /**
+   * The modal dialog's body
+   */
+  children: PropTypes.node,
+  /**
+   * Title text of content
+   */
+  title: PropTypes.node,
```
maximuk: please sort

components/selectionBlock/selectionBlock.css
```diff
@@ -66,6 +66,7 @@
   justify-content: center;
   margin: 0;
   width: 290px;
+  min-width: 290px;
```
maximuk: please sort

components/modal/modalContent.js
```diff
@@ -0,0 +1,38 @@
...
+  /**
+   * Set of custom modifications.
+   */
+  mods: PropTypes.arrayOf(PropTypes.string),
```
iwwwi: pls do not add `mods` because those are planned to be deprecated.. better do add `classname` if is needed at all... (i do not see where you use the `mods` in this component

components/modal/modal.js
```diff
@@ -256,6 +258,10 @@ Modal.propTypes = {
    * Determine whether a overlay is visible or not
    */
   overlay: PropTypes.bool,
+  /**
+   * Modal size
+   */
+  size: PropTypes.oneOf(['small', 'big']),
```
iwwwi: Since we have only 2 possible choices is it better to go with `true/false` variant? In that case we won't have unused class like - `size_big`

EduardTrutsyk:
maybe make sense to add `className` for `Modal` and for `ModalContent`?

components/modal/modal.js
```diff
@@ -200,6 +204,7 @@ Modal.defaultProps = {
   onClose: null,
   onOverlayClick: null,
   overlay: true,
+  size: false,
```
froskie: Passing `false` is odd here, isn't it? Wouldn't `null` be a better default?

froskie: Also, where is this being used?

spzm: There are two sizes of modal available: big (max-width: 780), small (max-width: 480).
If I specify size as Boolean I prefer set it as *false* instead of *null* value.

iwwwi: I guess here the prop should be `isSmall` and then will have the default `false`

components/modal/modal.scss
```diff
@@ -114,31 +113,48 @@
 }
 
 .ui-modal__header {
+  align-items: center;
   border-bottom: 2px solid var(--tx-modal-content-devider-color);
   color: var(--tx-modal-color-dark);
+  display: flex;
+
```
iwwwi: remove empty line pls

components/modal/modal.js
spzm: yep, thanks, fixing

components/selectionBlock/selectionBlock.css
```diff
@@ -65,6 +65,7 @@
   display: flex;
   justify-content: center;
   margin: 0;
+  min-width: 290px;
```
asci: why is this required?

spzm: Ah, forgot to mention in description, small fix for selection block.

Issue #325, opened 2018-02-05
Remove padding in SlidingPanel


ilabacheuski:
# Description

That is proposal to remove padding from Sliding Panel component. In production sometimes you do not need any padding. for an example if content comes from third party.
That would be nice to remove padding at least for content in https://github.com/Travix-International/travix-ui-kit/blob/master/components/slidingPanel/slidingPanel.scss#L58

PR #330, opened 2018-02-08
Remove padding from slidding panel content section

ilabacheuski:
# What does this PR do:

Remove padding from sliding Panel content section
"closes:#325"

diffs


yurist38:
Yeah, I remember that issue, I made a workaround for that once. Here I do not agree that it is a `patch` because once it merged it will affect all sliding panels across all apps, we have at least check them to be sure that the change is not breaking their layout. But I agree that we should remove this padding.

iwwwi:
1. I am interested in the cases where you need this padding vs when you don't need it;
2. the strategy to add the `padding: 20px` when you add the `padding: 0` for all cases when this padding is needed;
3. is it according the designs we have?

ilabacheuski:
We do not need 20 px at all, since cars, seat map, hotels just fix this to 0 or have broken design. Moreover it is mistake to dictate design to the code you don't own.

yurist38:
So what is our decision on this PR? It's definitely minor change, breaking layout most probably. But I don't see a big problem to fix our implementations once this PR is released. We used sliding panel just in few places, for now. It will be easily fixed (in case it breaks styles). So I vote for merging this and notify all teams to check/fix their implementations of `slidingPanel`. I agree that there is no point to keep this padding in `ui-kit`'s codebase. It's just adding us (developers) problems.

froskie:
Although I agree that components within UI Kit shouldn't have their **position** changed in their own styles but within their container, this is tricky. Considering how we have our themes and their based on our design, they should follow a convention of padding/margins. Think about all the usages now will have to override the **unexposed** CSS classes…

ilabacheuski:
Let me ask @all Do you know where do we use it right now with paddings? With this PR Hotels are fixed to design that was initially approved at first implementation. SeatMap does not suffer and as I remember SeatMap overrides padding to 0. CarWidget is fixed for a better look. Do we have any other places that this PR can break? As I see, everyone tries to make `padding = 0` to avoid strange decision to make it 20px. I think it is much better to fix it now, then later, because we will have better ui-kit with 0 paddings, instead of ui-kit that we still override to 0 px.

asci:
If no objections I'll merge this PR tomorrow morning

yurist38:
BTW, since the PR was created from forked repo it can't create an acceptance link automatically. Please next time create it on your side running `npm run deploy:accp` and share the link in PR description. Thanks!

leandrooriente:
It'll not break Seat Map because we reset it to 0.

```css
.container :global(.ui-sliding-panel__content) {
  padding: 0;
}
```

yurist38:
Just had a talk with @iwwwi regarding this PR. Seems like there is a better way to achieve a consistency with paddings in slidingPanel. And considering that we don't have a variable for padding (that I assumed existing) I agree to keep the fixed (non-zero) value here. This still should be refactored but in a bit more complex way since we have different use cases for sliding panel including 2 different paddings. So I think it should be done in a PR which will include all possible cases and have a predefined padding. Agree that it's the way to achieve a consistency.

froskie:
Closed in favor of https://github.com/Travix-International/travix-ui-kit/pull/358

PR #335, opened 2018-02-13
fixed & improved styles for modal box (according to the design)

MadinaShad:
# What does this PR do:

    - Fixes styles issue on IE 11
was:    
image 2018-02-07 at 3 34 50 pm

now:
screen shot 2018-02-15 at 15 50 40

    - Fixes styles issue on safari 9/10
was:
screen shot 2018-02-15 at 12 00 37

now:
screen shot 2018-02-15 at 15 41 16

   - Aligns height of content according to the design (small screen)
was: 
old

now: 
new

### DIFF
--- a/components/modal/modalContent.css
+++ b/components/modal/modalContent.css
@@ -3,6 +3,7 @@
   border-radius: var(--tx-generic-border-radius);
   color: var(--tx-generic-color-text-dark);
   font-family: var(--tx-generic-font-secondary-font-family), var(--tx-generic-font-secondary-generic-family);
+  flex: 1 0 auto;
 }
 
 .ui-modal-content-block__title {
--- a/components/modal/modalContent.css
+++ b/components/modal/modalContent.css
@@ -40,10 +40,6 @@
   border-radius: 0;
 }
 
-.ui-modal-content-block:last-of-type {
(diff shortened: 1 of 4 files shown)


components/modal/modalContent.css
```diff
@@ -40,10 +41,6 @@
   border-radius: 0;
 }
 
-.ui-modal-content-block:last-of-type {
-  height: 100%;
```
spzm: This increase height of last content block to fill left part modal window. Could you please check solution with adding `min-heigth` instead of removing `height`?

MadinaShad: fill left part modal window --> where can i check it? The height of the last block is stretched to the bottom, but it should be stretch by content inside

spzm: hmm, this stretching made me confused. Because on some design we have it stretched to the bottom but on some to content inside. So, if we are going to content inside version - I'm ok with it :)

iwwwi: @spzm we discussed with Gerard and the modal should wrap the content inside unlike the sidePanel. It is easy to get confused, and was not very clear on the designs also... Now the designs are updated, especially for small states.

PR #337, opened 2018-02-22
Improve handleClose method to use event

zanovski:
# What does this PR do:

Needed different behavior for cases when user will click on the cross button and on submit button. For now, I'm not able to distinguish this clicks. Because clicking on any close button or changing component's props will trigger handleClose method and onTryingToClose callback and there are no differences between this calls.

One way, which I found, is to pass the event to handleClose in case of interacting with any DOM element. In case of props are changed arguments will be empty.

diffs


asci:
Would be nice to update docs also

PR #338, opened 2018-02-22
Extend Badge component

### LINKED ISSUE #329: Improve Badge component to support other positions (i.e. centered ones)
# Description

> We stuck with a case when `Badge` is centered on top. So I think Badge component should be improved to support all possible positions. I think it should support centered case on every side and corners as well for each side. Feel free to put your ideas if have any...

yurist38:
# What does this PR do:
* Closes #329 
* Introduced new prop `align` into `<Badge />` component. It accepts one of these 3 values: [`start`, `center`, `end`]. The default behavior is not changed and this new prop is optional.
* Markdown of the `<Badge />` component is refactored to be more user-friendly and nice. 😄 
* Made badge is more responsive: now left and right badges have limitation by width and will not exceed 100% of the container width. Also, arrows were improved and became really responsive (this approach already used in `Hotels` widget and the solution is stolen from there. 😄 This is how it looks now if we have a multiline badge:
screen shot 2018-02-22 at 17 17 17

* Check the diffs and test an acceptance link.

# Unit tests:
* Added

### DIFF
--- a/components/badge/badge.md
+++ b/components/badge/badge.md
@@ -1,44 +1,47 @@
 Position top:
 
     initialState = {
-      isVisibleTop: false,
-      isVisibleRight: true,
-      isVisibleBottom: true,
-      isVisibleLeft: true,
+      position: 'top',
       isArrow: false,
       isBorder: true,
+      isCentered: false,
       isColorBackground: false,
+      labelText: 'Label'
     };
 
     <div>
-      <Checkbox
-        checked={state.isVisibleTop}
-        name="isVisibleTop"
-        onChange={() => setState({ isVisibleTop: !state.isVisibleTop })}
+      <RadioButton
+        checked={state.position === 'top'}
+        id="position-top"
+        name="position"
+        onChange={() => setState({ position: 'top' })}
       >
-        top
-      </Checkbox>
-      <Checkbox
-        checked={state.isVisibleRight}
-        name="isVisibleRight"
-        onChange={() => setState({ isVisibleRight: !state.isVisibleRight })}
+        Top
+      </RadioButton>
+      <RadioButton
+        checked={state.position === 'bottom'}
+        id="position-bottom"
+        name="position"
+        onChange={() => setState({ position: 'bottom' })}
       >
-        right
-      </Checkbox>
-      <Checkbox
-        checked={state.isVisibleBottom}
-        name="isVisibleBottom"
-        onChange={() => setState({ isVisibleBottom: !state.isVisibleBottom })}
+        Bottom
+      </RadioButton>
+      <RadioButton
+        checked={state.position === 'left'}
+        id="position-left"
+        name="position"
+        onChange={() => setState({ position: 'left' })}
       >
-        bottom
-      </Checkbox>
-      <Checkbox
-        checked={state.isVisibleLeft}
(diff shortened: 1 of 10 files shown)


maximuk:
Better to have one property `align="left|center|right"`

components/badge/badge.md
```diff
@@ -1,78 +1,132 @@
...
     initialState = {
-      isVisibleTop: false,
-      isVisibleRight: true,
-      isVisibleBottom: true,
-      isVisibleLeft: true,
+      align: 'start',
+      position: 'top',
       isArrow: false,
       isBorder: true,
+      isCentered: false,
```
maximuk: please remove the line

components/badge/badge.md
```diff
@@ -1,78 +1,132 @@
...
       isColorBackground: false,
+      isReversed: false,
```
maximuk: please remove the line

components/badge/badge.md
leandrooriente: Still valid?

components/badge/badge.md
leandrooriente: Still valid?

tests/unit/badge/badge.spec.js
```diff
@@ -88,4 +88,44 @@ describe('Badge', () => {
 
     expect(wrapper).toMatchSnapshot();
   });
+
+  it('should render badge with "start" alignment', () => {
```
leandrooriente: Should we have a combined (position + align) test?

components/badge/badge.md
yurist38: Ah, crap. Removing. Thanks!

tests/unit/badge/badge.spec.js
yurist38: @leandrooriente good idea, will add now

AlleeX:
Hmm..

selection_266

yurist38:
@AlleeX that's interesting. Let me check it.

yurist38:
@AlleeX could you please check now? Should be slightly better. I found this problem explained here: https://stackoverflow.com/a/33094994

package.json
```diff
@@ -17,6 +18,7 @@
     "prebuild": "babel --copy-files ./components --out-dir lib --ignore *.scss,*.md &",
     "styleguide-build": "styleguidist build",
     "styleguide-server": "styleguidist server",
+    "styleguide:dev": "npm run styleguide-build && npm run styleguide-server",
```
iwwwi: 👍

package.json
```diff
@@ -5,6 +5,7 @@
   "main": "lib/index.js",
   "scripts": {
     "build:watch": "node ./builder/program -w",
+    "build:watch:theme": "node builder/program.js -t ./themes/_default.yaml -w",
```
iwwwi: 👍

package.json
froskie: I'd like to namespace everything in the same way. I really like the combination of the other two, so let's try to make them all the same? Change the `-` to `:` in the other two?

package.json
froskie: I didn't see the point of this one. This is what happens by default, If you look at the `builder.js` file.

PR #339, opened 2018-02-26
fixed styles for the list component

MadinaShad:
# What does this PR do:

    Fixes styles for the List component:
   1.)  aligns multi-line text after icons
screenshot-1
   2.) removes icons if the prop hideBullets is true
52d38c9b0015c9843ef3dbee7cf0aabe _image 2018-02-26 at 2 02 49 pm

# How it looks now:
screen shot 2018-02-26 at 15 09 38

    Diff/acceptance link

### DIFF
--- a/components/dropDown/dropDown.md
+++ b/components/dropDown/dropDown.md
@@ -105,7 +105,7 @@ DropDown with filter mode:
     initialState = {filterOptions1: [
       { label: 'One', value: 'One', disabled: true },
       { label: 'Two', value: 'Two', checked: true },
-      { label: 'Three', value: 'Three' },
+      { label: 'Very very very very very long option', value: 'Three' },
       { label: 'Four', value: 'Four' },
     ], filterOptions2: [
       { label: 'One', value: '1' },
--- a/components/dropDown/dropDown.scss
+++ b/components/dropDown/dropDown.scss
@@ -295,7 +295,6 @@ $select-arrow-color: var(--tx-dropdown-arrow-color);
     padding: 5px 15px 5px 20px;
 
     .ui-checkbox__text {
-      height: 16px;
       margin-top: 0;
     }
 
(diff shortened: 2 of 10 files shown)


AlexDudar:
if I apply `hideBullets` to the horizontal list, it becomes vertical.. not sure if it's desired behavior 

screen shot 2018-02-26 at 3 45 22 pm

MadinaShad:
@AlexDudar mmm, looks like we don't use the icons on horizontal list. That means we shouldn't use any props to remove them. No icons, no problem 😄

MadinaShad:
@leandrooriente oh, will check it, thanks!

AlexDudar:
@MadinaShad , technically icons are present, they just hidden since `li` having `display: inline-block`. I don't think that hiding icons should affect/check if list is vertical or horizontal, it's simply weird to me, but of course it's your call ;)

yurist38:
Actually, I have used a horizontal list with the icons on Seatmap. I guess it's still there (below the header of the sidepanel).

MadinaShad:
@AlexDudar @yurist38 icons are present ONLY for vertical list, if we need to show icons on horizontal list we have to extend our ui-kit) And that problem which Alex has mentioned before I will fix

EduardTrutsyk:
is it possible to use different icons at the same time?
Like this:
screenshot at 11-09-22
screenshot at 11-15-35

MadinaShad:
@EduardTrutsyk It’s a very good question. The list doesn’t have this functionality. We have to think about it later

EduardTrutsyk:
one more question, why the prop name is `hideBullets`? 
maybe better to name it `hideIcons`? 
and on you example we have check marks :smiley:

rbardini:
Agree with @EduardTrutsyk, I think we should have a more generic prop name for this. The [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-type) call them "markers". Maybe we could have a `noMarkers` prop?

iwwwi:
@EduardTrutsyk @rbardini can you pls create a proposal for extending this component with this options? Pls add those details in both github issues section and our JIra board :) tnx

PR #344, opened 2018-03-02
Add support to JSX on ToggleButton

### LINKED ISSUE #343: Allow toggleButton to receive a JSX as items property
# Description

Currently `toggleButton` only accepts strings as `items` property. Sometimes the designer asks for an icon or a specific part of the text in bold, so it would be super useful if we can pass JSX as props as well.

It'll also help when you and to pass a translated text using `<L10ns />` instead of use `l` function to return a string.

# Proposal

```js
<ToggleButton
  handleSelect={() => {}}
  selectedIndex={0}
  items={[
      <L10ns name={key1} />,
      <L10ns name={key1} />,
  ]}
/>

<ToggleButton
  handleSelect={() => {}}
  selectedIndex={0}
  items={[
      'Text 1',
      'Text 2',
  ]}
/>

<ToggleButton
  handleSelect={() => {}}
  selectedIndex={0}
  items={[
      <span>Hello <strong>world</strong></span>,
      <span>Bye <strong>world</strong></span>,
  ]}
/>

<ToggleButton
  handleSelect={() => {}}
  selectedIndex={0}
  items={[
      <span><img src={airplane.svg} /> Flights</span>,
      <span><img src={account.svg} /> My account</span>,
  ]}
/>
```

leandrooriente:
# What does this PR do:

Add a new component to UI Kit called `ToggleItem` who should be passed as children to `ToggleButton` with the content that you want instead of pass it as items property.

### Before
```js
<ToggleButton
  items=['First Item', 'Second Item', 'Third item']
  selectedIndex={0}
  handleSelect={selectItem}
/>
```

### After
```js
<ToggleButton selectedIndex={0} handleSelect={selectItem}>
  <ToggleItem>First Item</ToggleItem>
  <ToggleItem>Second Item</ToggleItem>
  <ToggleItem>Third Item</ToggleItem>
</ToggleButton>
```

### Another possibilities
```js
<ToggleButton selectedIndex={0} handleSelect={selectItem}>
  <ToggleItem><Icon />First Item</ToggleItem>
  <ToggleItem>Second <strong>Item</strong></ToggleItem>
  <ToggleItem><L10ns name={'third-item'} /></ToggleItem>
</ToggleButton>
```

The previous implementation will continue working but returning the `ToggleItem` implicitly and with a deprecation warning.

This component has only been used on Seat Map widget repository as you can see here: https://bitbucket.org/search?q=ToggleButton%20travix-ui-kit&account=%7B8e35b3c6-9071-4267-a027-b8ec544f88f3%7D which is also maintained by our team.

DIff

# Acceptance

- https://toggle-button-allow-jsx-ytntemqsqg.now.sh

Closes #343


leandrooriente:
Hey guys, I need some help to test it. 
For some reason, I'm having problems with the `event` validation. 
Probably something related to the `mount` method and events: https://github.com/airbnb/enzyme/issues/382

leandrooriente:
Sorry for the mess with the commit's author.  My username on bitbucket is `loriente` and on github is `leandrooriente` and I always forget to update the local config 😞

components/toggleButton/toggleButton.js
```diff
@@ -15,27 +18,52 @@ export default function ToggleButton(props) {
     selectedIndex,
   } = props;
 
-  if (!Array.isArray(items) || items.length < 2) {
+  if (
+    (!Array.isArray(items) || items.length < 2) &&
```
maximuk: you should also hide the component if it receives only one child

leandrooriente: Done

rbardini: Why do we render `null` if we have only one item/child? I know it is an edge case and that most of the time it doesn't make sense to have a toggle button with only one option, but isn't it more predictable to just render a single item than nothing at all?

leandrooriente: I don't think so. It's a toggle, to render it you should attend some minimum requirements otherwise you can just render a Button.

rbardini: I understand, but I also believe that code should make the least assumptions possible, unless it can break the implementation. Someone may want to dynamically build a toggle button and still show the option even if it is a single one (instead of handling it as a special case to render something else), or they can check if they only have a single item and not render the toggle button in the first place.

Either way, this is probably out of the scope of this PR 🙂

leandrooriente: Well, if you want to use a Toggle Button, it means that you want to give more than one option to the user. If for some reason you only have one option (even with dynamic data) you should select this option and only show a Label or something indicating that the only option was selected. 

This kind of logic/validation should not be here.

leandrooriente:
Just rebased with Travix master.

@spzm double border fixed 😄

leandrooriente:
@iwwwi all warnings fixed https://toggle-button-allow-jsx-ytntemqsqg.now.sh

leandrooriente:
Browser tested:
- IE 11
- Edge 16
- Chrome (Mac OS)
- Safari 11
- Safari 9

PR #345, opened 2018-03-07
Hot 1399 update buttons extend modal box

MadinaShad:
# What does this PR do:

   - Adds possibility in modal-box to add the description after subtitle

modal-box

    - Adds a arrow to the buttons

arrow1

arrows2

    acceptance link


iwwwi:
Failing tests. pls check it ;) https://cl.ly/0X3z0P1q3N16

MadinaShad:
@iwwwi opps, i forgot to update tests 😃 will fix it, thanks!

leandrooriente:
Is the arrow property really necessary? It seems to be a super specific case to be part of the UI kit.

Button component accepts a React Element as children, so if someone needs to add an arrow it's not that hard to implement.

MadinaShad:
@leandrooriente this will bring the buttons to a single standard, as they should look like according to the design, because currently you can find the wrong styles on different affiliates. It will speed up the development and easy to maintain. By using the ui-kit you know that the styles on all brands are already created and it is not necessary to recheck them

leandrooriente:
Perfect.

We have an exception about arrow colors on CheapTickets. Don't forget to double check it.

screen shot 2018-03-07 at 10 44 51

screen shot 2018-03-07 at 10 41 10

MadinaShad:
@leandrooriente it's updated already, thanks for caring ))

asci:
`arrow` prop name would probably confuse in future, regarding that in MyAccount we have back buttons with arrows in the different direction. What about `arrowRight`, `arrowLeft` ?

asci:
after rethinking the case I agree with @leandrooriente  we can consider this case as a separate entity (like "button with arrow"). For this `arrow` (or I prefer `withArrow`) property would suit well. 

Especially considering that for Vliegwinkel we will not show the actual arrow — this will be part of the theme. 

All the other implementations (like links with arrows, or arrows back, or other icons) should be done via passing childer icons when Icons initiative will be delivered.

Issue #296, opened 2017-12-27
Fix overflow for Modal in IE11


yurist38:
# Browsers and versions affected

> Our lovely IE11

# Description

> There is an issue with overflow content in Modal component

<img width="592" alt="screen shot 2017-12-27 at 10 52 16" src="https://user-images.githubusercontent.com/10692413/34378218-17aac19e-eaf5-11e7-9e10-088c2229c468.png">

# Steps to reproduce

> Open ui-kit site in IE11 and navigate to Modal component. Then click on `Open Custom Modal` button. You will see that it behaves differently compare to other browsers (bottom part of modalbox went below the screen).

# Expected results

> Same behavior as in other browsers - you should be able to see the whole modalbox and its central (content) part should be scrollable.

leandrooriente:
Closed https://github.com/Travix-International/travix-ui-kit/pull/347

PR #347, opened 2018-03-08
fix overflow for ie11

iwwwi:
# What does this PR do:

   Fixes overflow problem on IE 11 Issue #296 

before: 
https://cl.ly/3F0O1e0u092M

After:
https://cl.ly/3h0f3b0O3Q2S

    Diff in styles


froskie:
What's with the difference in the snapshots since you didn't touch any code related to that?

leandrooriente:
In IE 11 the modal title is scrolling together with the content.

The same behavior doesn't happen in Edge or Chrome for example.

froskie:
Although I agree with @leandrooriente , I believe this fix the overflow problem _per se_. We should address the fixed header and footer in somewhere else.

iwwwi:
@leandrooriente  IMHO at the moment users are blocked to scroll the content in general without this fix :D and the header not being there is not a blocker :D can be fixed later... but I do not have time now to check this at moment... :D we can either merge this PR as fix for this bug (being blocked to scroll) and create another issue to improve the header experience ... or decline everything until I get the time to work on this again... or someone else can extend this PR :D I am ok in general with all approaches :)

yurist38:
Header/footer problem is not fully a scope of the current PR as I can see. Agree that ideally to fix it here as well. But since it requires more resources, I prefer to merge this PR asap and create an issue for the header/footer.

leandrooriente:
👍 

I created a new issue to track this problem: https://github.com/Travix-International/travix-ui-kit/issues/353

PR #354, opened 2018-03-13
Update package-lock and snapshots

yurist38:
# What does this PR do:

* Since we have this weird situation https://github.com/adriantoine/enzyme-to-json/issues/98 we have to update package lock to include last published version of that package
* Added `npmrc` setting to force the repo use package-lock

* Diff


yurist38:
Just updated readme file to use `npm ci` command instead

yurist38:
I had to update snapshots again. In this time I changed only one specific package version but tests still failed.

rbardini:
The dependency is directly related to snapshot creation, and the maintainers pulled the version because it had [breaking changes](https://github.com/adriantoine/enzyme-to-json/pull/91). That's probably why we have to update the snapshots again 🙂

PR #359, opened 2018-03-18
Remove wrong persist on markdown file

### LINKED ISSUE #351: "TypeError: e.persist is not a function" when pressing "Esc" while modal open
> All browsers

# Description
> After opening the modal and pressing "Esc" key, modal will throw error.

image

# Steps to reproduce
> 1 - Open any modal
> 2 - Hit extra keys

# Expected results
> Modal closes with no error

# Actual results
> Error is throw

leandrooriente:
#### What does this PR do:

Fix #351

### DIFF
--- a/components/modal/modal.md
+++ b/components/modal/modal.md
@@ -130,8 +130,7 @@ Basic Modal:
             <Modal
               active={state.isOpenModalWithCard}
               isSmall
-              onClose={(e) => {
-                e.persist();
+              onClose={() => {
                 setState({ isOpenModalWithCard: false });
               }}>
                 Modal Content


components/modal/modal.md
```diff
@@ -130,8 +130,7 @@ Basic Modal:
             <Modal
               active={state.isOpenModalWithCard}
               isSmall
-              onClose={(e) => {
-                e.persist();
```
AlleeX: Probably will be better to use like this? `e && e.persist();`

leandrooriente: But why do we need this?

AlleeX: Checked. For now it works fine without persist(). Thanks ;)

